PR #16, opened 2021-11-22
feat: add trojan source detection

### LINKED ISSUE #15: Add anti-trojan detection
The goal of this task is to add a new warning to signal malicious characters in source code.

If you don't know what I'm talking about yet, here is some reading:
- https://www.trojansource.codes/

The verification itself is not that hard and can be inspired by the following code: https://github.com/lirantal/anti-trojan-source/blob/main/src/main.js#L4

Because the whole string is required for the analysis, i guess the code is going to be somewhere here: https://github.com/NodeSecure/js-x-ray/blob/master/index.js#L13

antoine-coulon:
Fixes #15 

Standard implementation of the trojan source detection using a list of dangerous Unicode control characters (see more here : [https://www.unicode.org/reports/tr9/](url)).

For now, a warning is collected whenever one of the unsafe control char is found in a source string. Here is an example of the collected warning : 
`{
  kind: 'obfuscated-code',
  location: [ [ 0, 0 ], [ 0, 0 ] ],
  value: 'trojan-source'
}`

The analysis is done directly in a brand new Analysis class method `analyzeSourceString()` that could be used for all checks upstream the Tree walk requiring the whole source string.

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,6 +21,7 @@ export function runASTAnalysis(str, options = Object.create(null)) {
   });
 
   const sastAnalysis = new Analysis();
+  sastAnalysis.analyzeSourceString(str);
 
   // we walk each AST Nodes, this is a purely synchronous I/O
   walk(body, {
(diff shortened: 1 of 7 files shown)


fraxken:
@all-contributors please add @antoine-coulon for code

PR #23, opened 2022-04-27
Enhance exported warnings to include i18n token

Mathieuka:


index.js
```diff
@@ -85,14 +85,41 @@ export async function runASTAnalysisOnFile(pathToFile, options = {}) {
...
-    unsafeRegex: "unsafe-regex",
-    unsafeStmt: "unsafe-stmt",
-    unsafeAssign: "unsafe-assign",
-    encodedLiteral: "encoded-literal",
-    shortIdentifiers: "short-identifiers",
-    suspiciousLiteral: "suspicious-literal",
-    obfuscatedCode: "obfuscated-code"
+    parsingError: {
+      code: "ast-error",
+      i18n: "ast_error"
```
fraxken: `warnings.ast_error`. You need the full i18n path here (same for all the i18n properties below).

fraxken (changes requested):
You also need to update the TypeScript definition file:
https://github.com/NodeSecure/js-x-ray/blob/f833d69b3f692a2c87ab3c48067a425dae76e14d/index.d.ts#L57-L67

Maybe we can also add a section in README to explain how we can use CONSTANTS warnings with the i18n code ?

README.md
```diff
@@ -73,6 +73,44 @@ The analysis will return: `http` (in try), `crypto`, `util` and `fs`.
...
+We make a call to `i18n` through the package `NodeSecure/i18n` to get the translation.
+```
+import * as scanner from "@nodesecure/scanner";
+import * as i18n from "@nodesecure/i18n";
+
+const payload = await scanner.from("packageName", { 4 }, initLogger("packageName"));
+
+if (payload.warnings.length) {
+  const value = i18n.getToken(payload.warnings[0].i18n);
+}
```
fraxken:
```js
import * as jsxray from "@nodesecure/js-x-ray";
import * as i18n from "@nodesecure/i18n";

console.log(i18n.getToken(jsxray.CONSTANTS.Warnings.parsingError.i18n));
```

An example like this is better (payload.warnings is not about JS-X-Ray warnings).

Mathieuka: Yes

index.js
```diff
@@ -85,14 +85,41 @@ export async function runASTAnalysisOnFile(pathToFile, options = {}) {
 
 export const CONSTANTS = {
```
Kawacrepe: Do we want to keep exporting `export const CONSTANTS = {
  Warnings: ...
}`

Instead exporting directly the warnings object

cf: our discussion on discord #js-x-ray chan

Mathieuka: Yes !

fraxken (approved):
LGTM

Kawacrepe (approved):
neat!

Issue #25, opened 2022-05-06
Detect and throw warning for weak crypto hash algorithm


fraxken:
The goal of this task (issue) is to develop a new feature capable of detecting any usage of weak hash algorithm like `md5`.

For the sake of simplicity it is sufficient to look for the createHash method.

Example of code that should throw a new warning:
```js
import crypto from "crypto";

crypto.createHash("md5");
```

We may have to answer few questions for this issue:

- Do we have to handle other **API** (like the [WebCrypto API](https://nodejs.org/api/webcrypto.html))? Maybe we can also in some ways deal with popular crypto library ?
- Is there is another algorithms that we are considering "weak" other than md5 ? (i guess `sha1` has to be considered weak too).

fraxken:
@Mathieuka was also interested by this one (so maybe check with him if he still want to do it).

Mathieuka:
@fraxken Finally I don't feel strong enough to take this ticket at the moment, I'm going to keep learning about the AST part :muscle: 

@tony-go If you do a live on this ticket Tony that would be great !!

fraxken:
@all-contributors please add @tony-go for code, doc, test

PR #30, opened 2022-05-24
feat: detect weak crypto algorithm

tony-go:
## Context

This PR resolves issue #25 

## Improvements 🚀 

At the moment we detect:

| Algorithm | Code              | import 'crypto' |
|-----------|-------------------|-----------------|
| `md5`     | createHash('md5') |  ✅              |
| `md5`     | crypto.createHash('md5') |  ✅              |

## Questions

- Should we make `import 'crypto'` be required to trigger this warning?
- Should I add other weak crypto algorithms in this PR? (like `sha1`, but TBH I don't know other ones 🙈)?


Kawacrepe (commented):
Lgtm! My only question is: shouldn't we add a valid test case ? where the cryptoHash algorithm used is something like sha256 etc..?

tony-go:
> Lgtm! My only question is: shouldn't we add a valid test case ? where the cryptoHash algorithm used is something like sha256 etc..?

Very good idea ^^ I'll add it.

Kawacrepe (approved):

tony-go:
@fraxken do you prefer to wait for the tracer feature?

fraxken:
@tony-go No i guess this is ok to merge. For weak hash i have found this list
- MD2, MD4, MD5, RIPEMD-160, and SHA-1

SHA1 is supported by Node.js (maybe we have to check for the other ones).

src/probes/isWeakCrypto.js
```diff
@@ -0,0 +1,26 @@
...
+    node.callee.property.name === "createHash";
+
+  return [isSimpleIdentifier || isMemberExpression];
+}
+
+function main(node, { analysis }) {
+  const arg = node.arguments.at(0);
+  if (arg.value === "md5" && analysis.dependencies.has("crypto")) {
+    analysis.addWarning(warnings.weakCrypto, "md5", node.loc);
+  }
```
fraxken: Make the code dynamic to check multiple hash algorithms. The list of vulnerable hashes could be stored in an ES6 Set.

index.js
```diff
@@ -128,6 +128,11 @@ export const warnings = Object.freeze({
     code: "obfuscated-code",
     i18n: "sast_warnings.obfuscated_code",
     severity: "Critical"
+  },
+  weakCrypto: {
+    code: "weak-crypto",
+    i18n: "sast_warnings.weak_crypto",
+    severity: "Information"
```
fraxken: Maybe we should add another property `experimental: true` here ? So other tools like CLI may use this to show an experimental note.

fraxken (changes requested):

src/probes/isWeakCrypto.js
```diff
@@ -0,0 +1,34 @@
+// Internal Dependencies
+import { warnings } from "../constants.js";
+
+// Constants
+const weakAlgorithms = ["md5", "sha1", "ripemd160", "md4", "md2"];
```
fraxken: `new Set(["md5", "sha1", "ripemd160", "md4", "md2"])` no ?

src/probes/isWeakCrypto.js
```diff
@@ -0,0 +1,34 @@
...
+
+  return [isSimpleIdentifier || isMemberExpression];
+}
+
+function main(node, { analysis }) {
+  const arg = node.arguments.at(0);
+  const isCryptoImported = analysis.dependencies.has("crypto");
+
+  if (
+    weakAlgorithms.includes(arg.value) &&
```
fraxken:
```js
weakAlgorithms.has(arg.value)
```

With an ES6 Set

src/probes/isWeakCrypto.js
tony-go: Shame on me! this is the vim effect!

src/probes/isWeakCrypto.js
tony-go: 🙈

fraxken:
@all-contributors please add @tony-go for code,tests,docs

fraxken:
@all-contributors please add @tony for code, doc, test

PR #35, opened 2022-06-08
test(probes): test unit of isArrayExpression probe

Mathieuka:


fraxken:
Closing the PR has this is now a bit out date and not updated

PR #57, opened 2022-10-29
refactor: implement new VariableTracer

fraxken:
This PR finalize the implementation of the new NodeSecure VariableTracer class. Two probes already use it:

- isRequire
- isWeakCrypto

A lot of code has been removed because the old mechanism was no more useful.

This PR also include a Major (breaking) changes by totally removing the `unsafe-assign` warning. This one was related to the old mechanism and I think most of the warnings were not relevant in the real world.

Because initial implementation were not capable to catch a lot of situations I added this warning but now the Tracer is quite capable. So I think removing it will greatly help our tools like CI and avoid to match warnings in the UI.


antoine-coulon (approved):
LGTM, great stuff :)

src/probes/isWeakCrypto.js
```diff
@@ -1,26 +1,30 @@
...
-const kWeakAlgorithms = new Set(["md5", "sha1", "ripemd160", "md4", "md2"]);
+const kWeakAlgorithms = new Set([
+  "md5",
+  "sha1",
+  "ripemd160",
+  "md4",
+  "md2"
+]);
+
+function validateNode(node, { tracer }) {
```
tony-go, line 13: ❤️

test/probes/weakCrypto.spec.js
```diff
@@ -49,8 +49,11 @@ test("it should report a warning in case of `[expression]createHash(<weak-algo>)
 });
 
 test("it should NOT report a warning in case of `[expression]createHash('sha256')` usage", (tape) => {
```
tony-go, line 48: I just like to have a test like:

```js
const md5Usage = readFileSync(join(FIXTURE_PATH, "strongAlgorithms", "sha256.js"), "utf-8");
const { warnings: outputWarnings } = runASTAnalysis(md5Usage);
const code = `
    import c from 'crypto';
    const crypto = {
      createHash() {}
    }
    crypto.createHash('sha1');
`;
const { warnings: outputWarnings } = runASTAnalysis(code);

tape.strictEqual(outputWarnings.length, 0);
tape.end();
```

It checks that even if I have `crypto` imported in the scope imported and I use a `createHash` from another object, it won't raise an error. For me, it's the case I was not capable to catch before the tracer.

tony-go (approved):
LGTM 🎉 

I'd prefer to see the test I mentioned before we merge this, but it's up to you. 🙌 

Thanks for the investment you put into that project. Can't wait to read more about the code in ast-utils.

PR #60, opened 2022-11-12
refactor: use file urls in tests

targos:


tony-go:
Thanks @targos :) let run tests and see

Issue #59, opened 2022-11-07
Parsing error: depName.trim is not a function


targos:
![image](https://user-images.githubusercontent.com/2352663/200280173-6b3d6e56-16b5-4ce4-8263-0e1a351e5e65.png)

fraxken:
Fixed with https://github.com/NodeSecure/js-x-ray/pull/65 (release coming soon).

PR #61, opened 2022-11-12
[wip] Fix depName.trim is not a function

targos:
Fixes: https://github.com/NodeSecure/js-x-ray/issues/59

### DIFF
--- a/test/fixtures/regress/prop-types.min.js
+++ b/test/fixtures/regress/prop-types.min.js
@@ -1,0 +1,26 @@
+/*
+prop-types module, version 15.8.1, licensed as:
+
+MIT License
+
+Copyright (c) 2013-present, Facebook, Inc.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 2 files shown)


targos:
@fraxken I marked it wip for now because I'm not sure how you would like this to be handled.

The problem is that `prop-types.min.js` is in a sort of UMD bundle that remaps internal `require` calls using numbers as module identifiers.

The builtin Node.js `require` function doesn't allow to pass something that isn't a string, so we may be safe if we just ignore calls with a literal that isn't a string?

Here is the relevant code:

https://github.com/NodeSecure/js-x-ray/blob/cddf98856fec6c938ca4da658bcdf32701de1b51/src/probes/isRequire.js#L71-L74

```
{
  arg: {
    type: 'Literal',
    value: 3,
    raw: '3',
    loc: { start: [Object], end: [Object] }
  },
  node: {
    type: 'CallExpression',
    callee: { type: 'Identifier', name: 'require', loc: [Object] },
    arguments: [ [Object] ],
    loc: { start: [Object], end: [Object] }
  }
}
```

targos:
Feel free to push commits to this branch.

targos:
@tony-go maybe you can help?

tony-go:
hey @targos 👋 

Happy to see you again 💪 

Meriyah parser didn't differentiate string literal from numeric literal (like babel parser does) - So yeah, you should add a control flow statement before pushing it into deps.

-------------------------------
> The builtin Node.js require function doesn't allow to pass something that isn't a string,

Now I'm asking: should we have another prob to catch it and raise it as a warning? WDYT

In other terms: In the context of node, you don't want to let something like this happen? Am I correct?

fraxken:
Sorry, I was sick for a whole week so I didn't had much time to dedicate on this issue!

---

> Now I'm asking: should we have another prob to catch it and raise it as a warning? WDYT

I don't think adding a warning is a good idea because here we are aware that on Node executing `require(3)` will throw an ERR_INVALID_ARG_TYPE.

Except if there is a way to exploit this I don't see value in adding a warning that will only be triggered in similar case like UMD.

---

I think a good way to fix this is to return if `depName` is not a string in ASTDeps because I don't expect anything else than a string anyway.

```js
if (typeof depName !== "string" || depName.trim() === "") {
  return;
}
```

This will also cover other scenarios like `require(1+1)`, `require([1, 1])` etc..

PR #65, opened 2023-01-01
fix(ASTDeps): depName.trim is not a function

fraxken:


fraxken:
@all-contributors please add @targos for code, bug

targos:
Thanks!

PR #105, opened 2023-03-09
feat: add shady-link warning

### LINKED ISSUE #80: Detect shady links
My idea of this task is to implement a new warning responsible of detecting shady links in Literals. I have been inspired by one of the detection of [guarddog](https://github.com/DataDog/guarddog) from DataDog.

They use the following RegEx:
`(http[s]?:\/\/.*\.(link|xyz|tk|ml|ga|cf|gq|pw|top|club|mw|bd|ke|am|sbs|date|quest|cd|bid|cd|ws|icu|cam|uno|email|stream))`

This RegEx allow to detect an URL like that:

```
https://foobar.xyz
```

The idea is to create a new probe or to update isLiteral probe to add that new warning. The probe will execute on every ESTree Literal.

```
const maliciousUrl = "https://foobar.xyz";
```

---

Maybe we need to conduct additional research on the subject (maybe there is some study we may want to read to improve the detection?).

PierreDemailly:
This PR fixes #80

### DIFF
--- a/src/probes/isLiteral.js
+++ b/src/probes/isLiteral.js
@@ -6,6 +6,10 @@ import { Hex } from "@nodesecure/sec-literal";
 
 // CONSTANTS
 const kNodeDeps = new Set(builtinModules);
+const kShadyLinkRegExps = [
+  /(http[s]?:\/\/bit\.ly.*)$/,
+  /(http[s]?:\/\/.*\.(link|xyz|tk|ml|ga|cf|gq|pw|top|club|mw|bd|ke|am|sbs|date|quest|cd|bid|cd|ws|icu|cam|uno|email|stream))$/
+];
 
 /**
  * @description Search for Literal AST Node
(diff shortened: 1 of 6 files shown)


antoine-coulon (approved):
Nice one! LGTM

src/probes/isLiteral.js
```diff
@@ -37,6 +37,9 @@ function main(node, options) {
       analysis.addWarning("encoded-literal", node.value, node.loc);
     }
   }
+  else if (/(http[s]?:\/\/.*\.(link|xyz|tk|ml|ga|cf|gq|pw|top|club|mw|bd|ke|am|sbs|date|quest|cd|bid|cd|ws|icu|cam|uno|email|stream))$/.test(node.value)) {
```
Kawacrepe: If I understand correctly, on guarddog they also consider `(http[s]?:\/\/bit\.ly.*)$` as a shady link regex, shouldn't we add it?

tony-go: We could put the regex into a variable. It could improve readability.

tony-go (approved):
lgtm & my comment

src/probes/isLiteral.js
PierreDemailly: I think we should ;) It's done

PierreDemailly: I agree, it's done

Kawacrepe (approved):

fraxken (approved):
LGTM

fraxken:
@all-contributors please add @PierreDemailly for code, test

cccs-kevin:
This is awesome, I look forward to the next release that includes this!

PR #108, opened 2023-03-18
refactor: migrate to Node.js native test runner

fraxken:


test/index.js
PierreDemailly, line 1: We need this even with glob pattern ?

test/obfuscated.spec.js
```diff
@@ -12,84 +11,87 @@ import { getWarningKind } from "./utils/index.js";
...
-  tape.deepEqual(getWarningKind(warnings), ["obfuscated-code"].sort());
-  tape.strictEqual(warnings[0].value, "jsfuck");
-  tape.end();
+  assert.strictEqual(warnings.length, 1);
+  assert.deepEqual(getWarningKind(warnings), ["obfuscated-code"].sort());
+  assert.strictEqual(warnings[0].value, "jsfuck");
 });
 
-// test("should detect 'morse' obfuscation", (tape) => {
+// test("should detect 'morse' obfuscation", () => {
```
PierreDemailly, line 23: Should we delete or uncomment this test ?

test/index.js
fraxken: Yes because it will try to execute other files (like fixtures) that are not tests

test/obfuscated.spec.js
fraxken: https://github.com/NodeSecure/js-x-ray/issues/144 (out the scope of this PR)

PierreDemailly (approved):
LGTM
edit: % tape conflict

PR #160, opened 2023-11-25
Add a source property to Warnings

### LINKED ISSUE #159: Add a source property to Warnings
I would like to a new `source` property to warnings to be able to identity who created the warning. Nowadays warnings can be generated from JS-X-Ray or Scanner (take [invalid-semver warning](https://github.com/NodeSecure/scanner/blob/bf36fb0c44df63b00dc2662afe79aefaa0fd3689/src/utils/semver.js#L51)).

The task consist of updating the TS interface and add the source in the [generateWarning](https://github.com/NodeSecure/js-x-ray/blob/master/src/warnings.js#L56) function.

https://github.com/NodeSecure/js-x-ray/blob/c91a37e8e530b72320c497c2beae475b3d00415d/types/warnings.d.ts#L24-L32

fabnguess:
The purpose of this PR is to resolve #159 

close #159

### DIFF
--- a/src/warnings.js
+++ b/src/warnings.js
@@ -54,16 +54,16 @@ export const warnings = Object.freeze({
 });
 
 export function generateWarning(kind, options) {
-  const { location, file = null, value = null } = options;
+  const { location, file = null, value = null, source = "JS-X-Ray" } = options;
 
   if (kind === "encoded-literal") {
     return Object.assign(
-      { kind, value, location: [utils.toArrayLocation(location)] },
+      { kind, value, location: [utils.toArrayLocation(location)], source },
       warnings[kind]
     );
   }
 
-  const result = { kind, location: utils.toArrayLocation(location) };
+  const result = { kind, location: utils.toArrayLocation(location), source };
   if (utils.notNullOrUndefined(file)) {
     result.file = file;
   }
--- a/types/warnings.d.ts
+++ b/types/warnings.d.ts
@@ -25,6 +25,7 @@ interface WarningDefault {
   kind: WarningName;
   file?: string;
   value: string;
+  source: string; 
   location: WarningLocation | WarningLocation[];
   i18n: string;
   severity: "Information" | "Warning" | "Critical";
--- a/test/warnings.spec.js
+++ b/test/warnings.spec.js
@@ -14,6 +14,7 @@ test("Given an encoded-literal kind it should generate a warning with deep locat
   assert.deepEqual(result, {
     kind: "encoded-literal",
     value: null,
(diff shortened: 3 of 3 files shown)


src/warnings.js
```diff
@@ -54,12 +54,12 @@ export const warnings = Object.freeze({
 });
 
 export function generateWarning(kind, options) {
-  const { location, file = null, value = null } = options;
+  const { location, file = null, value = null, source = "JS-X-Ray"} = options;
```
PierreDemailly: Should we validate source ?
```js
const kWarningsSource = new Set(["JS-X-Ray", "Scanner"]);
// [...]
if (!kWarningsSource.has(source) {
  throw new Error(`Source ${source} is not allowed`);
}
```

fabnguess: Sorry, I hadn't seen that before my last commit. I've finally removed the union type (should I leave it in place?).

PierreDemailly: I think it's ok to not check the source then 👍 
However, i would not set it as `null` by default but `JS-X-Ray` instead.

fabnguess: OK noted

types/warnings.d.ts
```diff
@@ -25,6 +25,7 @@ interface WarningDefault {
   kind: WarningName;
   file?: string;
   value: string;
+  source?: string;
```
fraxken: should be a mandatory property

src/warnings.js
```diff
@@ -54,11 +54,11 @@ export const warnings = Object.freeze({
 });
 
 export function generateWarning(kind, options) {
-  const { location, file = null, value = null } = options;
+  const { location, file = null, value = null, source = "JS-X-Ray" } = options;
 
   if (kind === "encoded-literal") {
     return Object.assign(
-      { kind, value, location: [utils.toArrayLocation(location)] },
+      { kind, value, location: [utils.toArrayLocation(location)], source },
```
fraxken, line 61: You need to add source to the `result` Object too (few lines below).

fraxken (changes requested):

src/warnings.js
fraxken: I dont think we need to validate source (could limit customization of source in future).

src/warnings.js
PierreDemailly: @fabnguess you missed the `result` const line 66

PierreDemailly (approved):

fraxken (approved):

fraxken:
@allcontributors please add @fabnguess for code

PR #164, opened 2023-11-28
fix: illegal return statement parsing error

### LINKED ISSUE #163: SyntaxError: Illegal return statement
A static analysis of the following code: https://unpkg.com/which@3.0.0/bin/which.js throw `ParseError [SyntaxError]: [15:8]: Illegal return statement`

here is a minimal reproduction

```js
const argv = process.argv.slice(2);

function foobar() {
  console.log("foobar");
}

if (!argv.length) {
  return foobar();
}
```

This code is legal in Node.js CJS because the module is wrapped in function (I think).

fraxken:
fix #163

### DIFF
--- a/test/illegalReturnStatement.spec.js
+++ b/test/illegalReturnStatement.spec.js
@@ -1,0 +1,31 @@
+// Import Node.js Dependencies
+import { test } from "node:test";
+import assert from "node:assert";
+
+// Import Internal Dependencies
+import { runASTAnalysis } from "../index.js";
+
+/**
+ * @see https://github.com/NodeSecure/js-x-ray/issues/163
+ */
+// CONSTANTS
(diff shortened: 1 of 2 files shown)


test/illegalReturnStatement.spec.js
```diff
@@ -0,0 +1,27 @@
...
+  console.log("foobar");
+}
+
+if (!argv.length) {
+  return foobar();
+}
+`;
+
+test("it should not throw error whatever module is true or false", () => {
+  runASTAnalysis(kIncriminedCodeSample, { module: false });
```
PierreDemailly: Should we either use `assert.doesNotThrow` or remove assert import ?

PierreDemailly (approved):

PR #185, opened 2023-12-17
fix the example in readme to prevent "location" field displays wrong way when running as script with Node.js. (nested displayed as [Array])

zxkmm:
this PR fix the example in readme to prevent "location" field displays wrong way when running as script with Node.js. (nested displayed as [Array])  
Since this repo doesn't have a "dev" branch and i don't have permission to this repo to push more branches, this PR directly pointed to master/main branch. If i need to change it, feel free to ask me to do so.  
Thx

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,27 @@ console.log(dependenciesName);
 console.log(inTryDeps);
 console.log(warnings);
 ```
+---
+
+If you meant to run this example with Node.js quickly in terminal, use this example below to correctly display the ``location`` field.
+```js
+import { runASTAnalysis } from "@nodesecure/js-x-ray";
+import { readFileSync } from "fs";
+
+const str = readFileSync("./file.js", "utf-8");
+const { warnings, dependencies } = runASTAnalysis(str);
+
+const dependenciesName = [...dependencies];
+const inTryDeps = [...dependencies.getDependenciesInTryStatement()];
+
+console.log(dependenciesName);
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -80,6 +80,27 @@ console.log(dependenciesName);
 console.log(inTryDeps);
 console.log(warnings);
```
PierreDemailly: Hi @zxkmm, thanks for contributing!
I'm not fan to duplicate the whole example, we should directly update this console.log instead.

```suggestion
console.dir(warnings, { depth: null });
```

We can eventually add a note below the code snippet:

> [!NOTE]
> We use `console.dir` instead of `console.log` due to nested object.

WDYT ?

fraxken:
Thanks for opening a PR,

Not sure why adding a new example? We just need to update the one above no?

README.md
fraxken: No need to add any notes for me (like we are not going to explain every line of codes..)

zxkmm: Thank you so much! I didn't know this usage before you mentioned. Accepted this and commited.

fraxken (approved):

PierreDemailly (approved):

zxkmm:
Thank you so much for the approval.  :)

---


> We just need to update the one above no?      

I originally thought better to keep users have choices. But later @PierreDemailly nicely provided a nice usage I didn't know before.  Which works better. thx!  :)

Issue #179, opened 2023-12-03
Properly detect eval("require") as require


fraxken:
https://unpkg.com/file-type@16.5.4/core.js

We should detect the following code as an unsafe-import (but still detect the usage of the `stream` core module)

```js
// Using `eval` to work around issues when bundling with Webpack
const stream = eval('require')('stream'); // eslint-disable-line no-eval
```

fabnguess:
@fraxken, I can't find the instruction `const stream = eval('require')('stream')` anywhere in the source code. Could you please guide me?

PR #190, opened 2023-12-18
fix(unsafe-import): warning on unsafe-import using eval/require

tchapacan:
fix https://github.com/NodeSecure/js-x-ray/issues/179

### Warning on unsafe-import using eval/require

`const stream = eval('require')('stream')`

> should return unsafe-import & unsafe-eval ?

Purpose of this PR is to propose a fix for the issue https://github.com/NodeSecure/js-x-ray/issues/179 by mdofifying probes isRequire & isUnsafeCallee : 

- [x] modify probe isRequire
- [x] modify probe isUnsafeCallee
- [x] add tests
- [x] documentation

Feel free to update it or give any feedback i'll be happy to try to improve it (this is my first real PR 😄 i'm new to the environment so I didn't dig deeper yet sorry, but don't hesitate to give any suggestion to improve it)

Result after : 

```js
import { runASTAnalysis } from "./index.js";

const result = runASTAnalysis(`
const stream = eval('require')('stream')
`);

console.log(result);
```

```
➜  js-x-ray git:(fix/179) ✗ node test.js
{
  idsLengthAvg: 6,
  stringScore: 0,
  warnings: [
    {
      kind: 'unsafe-stmt',
      location: [Array],
      source: 'JS-X-Ray',
      value: 'eval',
      code: 'unsafe-stmt',
      i18n: 'sast_warnings.unsafe_stmt',
      severity: 'Warning'
    },
    {
      kind: 'unsafe-import',
      location: [Array],
      source: 'JS-X-Ray',
      value: 'stream',
      i18n: 'sast_warnings.unsafe_import',
      severity: 'Warning'
    }
  ],
  dependencies: Map(1) {
    'stream' => { unsafe: true, inTry: false, location: [Object] }
  },
  isOneLineRequire: true
}
```


src/utils.js
```diff
@@ -18,6 +18,22 @@ export function isUnsafeCallee(node) {
...
 
+export function isUnsafeConstEvalRequireImport(node) {
+  const isUnsafe = (node.declarations &&
+    node.declarations[0].init &&
+    node.declarations[0].init.callee &&
+    node.declarations[0].init.callee.callee &&
+    node.declarations[0].init.callee.callee.type === "Identifier" &&
+    node.declarations[0].init.callee.callee.name === "eval" &&
+    node.declarations[0].init.callee.arguments[0].type === "Literal" &&
+    node.declarations[0].init.callee.arguments[0].value === "require");
```
fraxken: You can use the following code to simplify

```js
const identifier = getCallExpressionIdentifier(node);
```

Then just need to assert that identifier is eval and doing check on `node.arguments.at(0)`

test/probes/isUnsafeImport.spec.js
```diff
@@ -0,0 +1,46 @@
...
+test("should not detect unsafe import", () => {
+  const str = "const stream = eval('fastify')('express');";
+
+  const ast = parseScript(str);
+  const sastAnalysis = getSastAnalysis(str, isUnsafeImport)
+    .execute(ast.body);
+
+  const result = sastAnalysis.getWarning(kWarningUnsafeImport);
+  assert.equal(result, undefined);
+});
```
fraxken: I think asserting the number of warnings in each tests could be a bit safer

src/utils.js
fraxken: And so I think we don't need to put in in utils

src/probes/index.js
```diff
@@ -15,9 +15,11 @@ import isUnaryExpression from "./isUnaryExpression.js";
 import isWeakCrypto from "./isWeakCrypto.js";
 import isClassDeclaration from "./isClassDeclaration.js";
 import isMethodDefinition from "./isMethodDefinition.js";
+import isUnsafeImport from "./isUnsafeImport.js";
```
fraxken: I would prefer isUnsafeEvalRequire

fraxken (changes requested):

src/probes/index.js
tchapacan: Make more sense totally agree, and it's fixed in latest commit

test/probes/isUnsafeImport.spec.js
tchapacan: Agree, it's fixed in latest commit. Some questions below: 

- shall we move the unsafe-stmt test somewhere else or it's good here ?
- i saw you added a folder for tests related to issues, shall we move it there? Or we can consider it as probe tests?

fraxken: Nop that's ok (issues could be useful if there is no other places).

src/probes/isUnsafeEvalRequire.js
```diff
@@ -0,0 +1,35 @@
...
+    isUnsafe && node.arguments.at(0).value
+  ];
+}
+
+function main(node, options) {
+  const { analysis, data: calleeName } = options;
+
+  analysis.addWarning("unsafe-import", calleeName, node.loc);
+
+  return Symbol.for("breakWalk");
```
fraxken: You need to remove that line (because the number of warnings in your test are wrong)

src/utils.js
tchapacan: I've tried to simplify it to maximum using the method you suggested me (thanks a lot), by keeping some (necessary?) conditional check to keep tests green, lmk if it seems correct to you

src/probes/isUnsafeEvalRequire.js
tchapacan: it's not necessary agree i've removed it in latest commit

src/probes/isRequire.js
```diff
@@ -11,28 +11,50 @@ import {
...
+    ];
+  }
 
   return [
-    data !== null && data.name === "require",
-    data?.identifierOrMemberExpr ?? void 0
+    false,
+    null
   ];
 }
```
fraxken: You don't need to update the original validateNode, just create two distinct functions. And at the bottom provide them to validateNode as Array.

tchapacan: thanks it should be ok in latest commit, two little point i wanted to highlight:

- I did a rebase in my fork from master into my branch as there has been some changes since last time, (but not sure if it is the clean way to do it), if this PR is a mess i can do an another one no worry
- I took the liberty to modify a little the `runOnProbes` in utils tests to support the array of function exported from the probe, lmk if that's ok

I still have to write the docs, i'm gonna do it this week

src/probes/isRequire.js
```diff
@@ -99,7 +119,7 @@ function main(node, options) {
       analysis.addWarning("unsafe-import", null, node.loc);
 
       // We skip walking the tree to avoid anymore warnings...
-      return ProbeSignals.Skip;
+      return Symbol.for("skipWalk");
```
fraxken: Need to keep the code with ProbeSignals here

src/probes/isRequire.js
```diff
@@ -28,14 +25,35 @@ function validateNode(node, { tracer }) {
   ];
 }
 
+function validateNodeEvalRequire(node, options) {
+  const id = getCallExpressionIdentifier(node);
+  const argument = getCallExpressionArguments(node);
```
fraxken: Move that line after the condition

src/probes/isRequire.js
```diff
@@ -28,14 +25,35 @@ function validateNode(node, { tracer }) {
...
+  if (id === null) {
+    return [false];
+  }
+
+  return [
+    id === "eval" && argument[0] === "require",
+    options?.identifiersName[0]?.name ?? void 0
```
fraxken: Use variable id here

src/probes/isRequire.js
```diff
@@ -28,14 +25,35 @@ function validateNode(node, { tracer }) {
...
   const { tracer } = analysis;
 
   if (node.arguments.length === 0) {
     return;
   }
   const arg = node.arguments.at(0);
+  const id = getCallExpressionIdentifier(node);
+
+
+  if (arg.value === "require" && id === "eval") {
```
fraxken: No need to reuse function here if calleeName already provide the information

src/probes/isUnsafeCallee.js
```diff
@@ -16,13 +15,9 @@ function main(node, options) {
...
-  return ProbeSignals.Skip;
 }
 
 export default {
   name: "isUnsafeCallee",
-  validateNode,
-  main,
-  breakOnMatch: false
+  validateNode, main, breakOnMatch: false
 };
```
fraxken: We should keep the original code. No need to remove the skip for me.

src/utils.js
```diff
@@ -13,7 +13,8 @@ export function isUnsafeCallee(node) {
   // For Function we are looking for this: `Function("...")();`
   // A double CallExpression
   return [
-    identifier === "eval" || (identifier === "Function" && node.callee.type === "CallExpression"),
+    (identifier === "eval" && node.callee.type === "Identifier") ||
```
fraxken: Not sure why we need to add `node.callee.type === "Identifier"` here?

src/probes/isRequire.js
```diff
@@ -163,8 +183,5 @@ function walkRequireCallExpression(nodeToWalk, tracer) {
 
 export default {
   name: "isRequire",
-  validateNode,
-  main,
-  breakOnMatch: true,
-  breakGroup: "import"
+  validateNode: [validateNodeRequire, validateNodeEvalRequire], main, breakOnMatch: true, breakGroup: "import"
```
fraxken: you can keep break lines for each properties

fraxken (changes requested):

src/probes/isUnsafeCallee.js
fraxken: Note that here we degrade this probe to solve an issue in another (not what we expect). For me we should move isRequire before this probe and then improve the break in isRequire

src/probes/isRequire.js
tchapacan: agree its fixed in latest commit

src/utils.js
tchapacan: i have removed this unecenssary part in latest commit (was trying to enforce some side case i was creating in my firsts attempts)

src/probes/isUnsafeCallee.js
tchapacan: agree i went back to the initial implementation of the probe and just moved the order by putting isRequire first in the  list.

src/probes/isRequire.js
tchapacan: agreed i have removed this unecessary function call

src/probes/isRequire.js
tchapacan: changed `options?.identifiersName[0]?.name` to `id` its more consistent with the behavior in the rest of the codebase and the dep is properly added using addDependency now

src/probes/isRequire.js
tchapacan: fixed it in latest commit

src/probes/isRequire.js
tchapacan: thanks its fixed in latest commit

fraxken (approved):

src/SourceFile.js
```diff
@@ -52,7 +52,7 @@ export class SourceFile {
     }
   }
 
-  addDependency(name, location = null, unsafe = false) {
+  addDependency(name, location = null, unsafe = this.dependencyAutoWarning) {
```
fraxken, line 56: Pense à déclarer la propriété sur la class tout haut (comme les autres)

avec dependencyAutoWarning = false

tchapacan: ca roule! its fixed

tchapacan:
I have updated the PR according to your comments, again **thanks a lot lot** for your help and your time, I couldn't have done it without your help ! 😄

fraxken (approved):

Issue #188, opened 2023-12-17
Remove mockedFunction for Node.js test runner mock fn


fraxken:
Use the Node.js test runner mock.fn instead of the current mockedFunction (we dit that because mock was not available at the time).

https://github.com/NodeSecure/js-x-ray/blob/54414390607d0e0a137c76e7e7d0cbe3aaad79a2/test/utils/index.js#L9-L27

PR #201, opened 2024-01-14
Remove mockedFunction for Node.js test runner mock method

jean-michelet:
This pull request addresses issue  [#188](https://github.com/NodeSecure/js-x-ray/issues/188): "Remove mockedFunction for Node.js test runner mock fn".

I observed that the issue was specifically related to the mocking of object methods. To address this, I have implemented the use of the `mock.method()` function instead of `mock.fn`.

I noticed that the file `isArrayExpression.js` was excluded from the test suite due to a missing `.spec` extension. I have corrected this by renaming the file to `isArrayExpression.spec.js`, ensuring that it is now included in the test suite.

Looking forward to your feedback and any further suggestions.


fraxken (approved):

fraxken:
@allcontributors please add @jean-michelet for test

PR #204, opened 2024-01-14
feat(probeRunner): assert probes method in proberunner

tchapacan:
// TODO: https://github.com/NodeSecure/js-x-ray/blob/master/src/ProbeRunner.js#L74

### Assert probes method in probeRunner constructor

Purpose of this PR is to : 

- [x] assert that validateNode is a function or array of functions
- [x] assert that main is a function
- [x] add unittest

Note that i have used `import assert from "node:assert";` in probeRunner class but same can be done using if conditional.
LMK if it seems ok and enough to cover this TODO in probeRunner class ? I can assert that name exist as well ?

If anyone want to participate/feedback/propose/improve solution don't hesitate!

### DIFF
--- a/src/ProbeRunner.js
+++ b/src/ProbeRunner.js
@@ -1,3 +1,6 @@
+// Import Native Dependencies
+import assert from "node:assert";
+
 // Import all the probes
 import isUnsafeCallee from "./probes/isUnsafeCallee.js";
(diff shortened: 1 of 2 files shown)


fraxken (approved):

fraxken (approved):

fraxken:
@allcontributors please add @tchapacan for code, test

Issue #168, opened 2023-11-29
fs-constants@1.0.0 index.js detected as minified


fraxken:
The following file is detected as minified: https://unpkg.com/fs-constants@1.0.0/index.js

expected: should not be detected as a minified file

jean-michelet:
Looking into it.

jean-michelet:
It comes from the library `is-minified-code`...

I logged here: https://github.com/NodeSecure/js-x-ray/blob/527319006e2b0ef337bde04bee8f7345585b63d4/index.js#L67

I think one line file expressions are considered minified by default.

Considered minified:
```js
module.exports = require('fs').constants || require('constants')
```

Considered NOT minified:
```js
module.exports = require('fs').constants 
|| require('constants')
```

Why is it a big deal?

fraxken:
@jean-michelet Nothing crazy, it just report the file as minified when not in the CLI (false positive).

Yes is-minified-code is not perfect, last time @PierreDemailly was sharing another package but it has downside too.

PierreDemailly:
Yeah, didn't find a more accurate lib unfortunately.
May we could check for a minimum length if the file is one line ? Something like `isMinified(str) && (str.split(/\r?\n/).length > 1 ? true : str.length > 200)` ?

PR #205, opened 2024-01-15
Report.isOneLineRequire should be true if single line LogicalExpression assignment

jean-michelet:
EDIT: After exchanges with @fraxken, we need to check the LogicalExpression (left, right) and `ConditionalExpression` (consequent, alternate).

---

This pull request addresses issue  [#168](https://github.com/NodeSecure/js-x-ray/issues/168).

I added an AST check to ensure that files containing only one line and composed of a `LogicalExpression` assignation, typically an export: `module.exports = require('fs') || require('constants')`, are considered as not minified.


tony-go (approved):
Looks good codewise.

Let @fraxken double-check, as I haven't put my nose into the code for a while.

PierreDemailly (approved):

index.js
```diff
@@ -46,7 +47,7 @@ export function runASTAnalysis(
   return {
     ...source.getResult(isMinified),
     dependencies: source.dependencies,
-    isOneLineRequire: body.length <= 1 && source.dependencies.size <= 1
+    isOneLineRequire: (body.length <= 1 && source.dependencies.size <= 1) || isOneLineExpressionExport(body)
```
fraxken:
```suggestion
    isOneLineRequire: body.length <= 1 && (source.dependencies.size <= 1 || isOneLineExpressionExport(body))
```

fraxken (approved):

PierreDemailly (approved):

Issue #178, opened 2023-12-03
evaluate path.join for require?


fraxken:
The aim would be to successfully resolve dependency paths like in the following code example;

```js
const bin = require(path.join('.', 'bin.js'))
```

For this given code we want:
- Normalized dependency with `./bin.js` path (normalized using UNIX path)
- No unsafe-import warning

There is possibility for having identifiers like `__dirname` to appear

```js
const bin = require(path.join(__dirname, 'bin.js'))
```

The challenge here is to compute the location of the file to dynamically inject the `__dirname` value (Maybe to handle in a second evolution).

fraxken:
@jean-michelet yes

jean-michelet:
Looking into it @fraxken

PR #212, opened 2024-01-19
build path.join called in require if args are string literals

jean-michelet:
This PR addresses partially issue [ #178](https://github.com/NodeSecure/js-x-ray/issues/178).

If `path.join` is called in `require` and every argument is a string literal:
```js
const bin = require(path.join('.', 'bin.js'))
```
The dependency is normalized an no unsafe-import warning is triggered.

### DIFF
--- a/src/probes/isRequire.js
+++ b/src/probes/isRequire.js
@@ -1,4 +1,6 @@
 /* eslint-disable consistent-return */
+// Import Node.js Dependencies
+import path from "node:path";
 
 // Import Third-party Dependencies
 import { Hex } from "@nodesecure/sec-literal";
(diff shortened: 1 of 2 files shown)


src/probes/isRequire.js
```diff
@@ -11,6 +11,7 @@ import {
   getCallExpressionArguments
 } from "@nodesecure/estree-ast-utils";
 import { ProbeSignals } from "../ProbeRunner.js";
+import path from "path";
```
fraxken: Voir message sur Discord

A mettre tout en haut du code

```js
import path from "node:path";
```

fraxken (changes requested):

fraxken (approved):

PierreDemailly (approved):

Issue #214, opened 2024-01-25
Customizable SourceParser


fraxken:
The goal of the task is to be able to provide a custom SourceParser to the runASTAnalysis function (if not provided it will take the default JsSourceParser for example).

The idea behind that is to allow anyone to extend/add a new Parsing mechanism (to support TypeScript source for example).

In my mind I see two build-in class:
- SourceParser (with the current constructor code)

https://github.com/NodeSecure/js-x-ray/blob/d26eafcf92d8da424bdf7cfdcbc590045d8527ee/src/SourceParser.js#L18-L36

- JsSourceParser extending SourceParser (with the current parseScript code). The default parser for JS-X-Ray.

https://github.com/NodeSecure/js-x-ray/blob/master/src/SourceParser.js#L50

If someone want to re-implement his own, it would look like this;

```ts
import { SourceParser, runASTAnalysis } from "@nodesecure/js-x-ray";
import { parse } from '@typescript-eslint/typescript-estree';

export class TsSourceParser extends SourceParser {
  parseScript() {
    const ast = parse(this.source, {});

    return ast;
  }
}

const { warnings, dependencies } = runASTAnalysis(
  readFileSync("./file.ts", "utf-8"),
  {
    sourceParser: TsSourceParser
  }
);
```

PR #216, opened 2024-01-27
Refactor runASTAnalysis functions to use class AstAnalyser

jean-michelet:
The aim of this PR is to make the project more flexible and allow users to add new behaviors more easily.

I've created an `AstAnalyser` class that implements the logic originally contained in `runASTAnalysis` and `runASTAnalysisOnFile`. The class can be extended and has a dependency on any EStree-compatible parser that respects the following abstraction:
```ts
interface SourceParser {
  parse(source: string, options: unknown): ESTree.Program;
}

interface AstAnalyser {
  constructor(parser: SourceParser): void;
  analyse: (str: string, options?: Omit<RuntimeOptions, "customParser">) => Report;
  analyzeFile(pathToFile: string, options?: Omit<RuntimeFileOptions, "customParser">): Promise<ReportOnFile>;
}
```
Example:
```js
const jsAnalyser = new AstAnalyser(new JsSourceParser());
const tsAnalyser = new AstAnalyser(new TsSourceParser());
```
I removed the  class `SourceParser` to favor ***composition over inheritance***. 
Responsibility for preparing source code (deleting html comments etc.) for analysis can be assigned to the new class `AstAnalyser`. It removes the need for this class from the workspaces, wich could be problematic to create the `TsSourceParser` ([214](https://github.com/NodeSecure/js-x-ray/issues/214)).

I kept the functions `runASTAnalysis` and `runASTAnalysisOnFile` so my additions extend the possibilities without breaking the existing. Users can pass a custom parser to these functions without having to completely modify their code:
```js
function runASTAnalysis(
  str,
  options = Object.create(null)
) {
  const {
    customParser = new JsSourceParser(),
    ...opts
  } = options;

  const analyser = new AstAnalyser(customParser);

  return analyser.analyse(str, opts);
}
```

I've moved the tests for these functions to the new `AstAnalyser` class and replaced them with integration tests verifying that the functions call the `AstAnalyser` methods with the right arguments.


types/api.d.ts
```diff
@@ -76,5 +73,15 @@ type ReportOnFile = {
   warnings: Warning[];
 }
 
+interface SourceParser {
+  parse(source: string, options: unknown): ESTree.Program;
+}
+
+interface AstAnalyser {
```
fraxken: declare class AstAnalyser plutôt non?

types/api.d.ts
```diff
@@ -76,5 +73,15 @@ type ReportOnFile = {
   warnings: Warning[];
 }
 
+interface SourceParser {
```
fraxken, line 76: declare class

src/AstAnalyser.js
```diff
@@ -0,0 +1,129 @@
+// Import Node.js Dependencies
+import fs from "fs/promises";
+import path from "path";
```
fraxken: You need to add `node:` prefix

src/AstAnalyser.js
```diff
@@ -0,0 +1,129 @@
...
+import isMinified from "is-minified-code";
+
+// Import Internal Dependencies
+import { SourceFile } from "./SourceFile.js";
+import { isOneLineExpressionExport } from "./utils/index.js";
+
+export class AstAnalyser {
+  /**
+   * @constructor
+   * @param { Parser } parser
```
fraxken: What is Parser type?

fraxken (changes requested):

src/AstAnalyser.js
```diff
@@ -1,6 +1,6 @@
 // Import Node.js Dependencies
-import fs from "fs/promises";
-import path from "path";
+import fs from "node:fs/promises";
+import path from "node:path";
```
jean-michelet, line 3: Ooops @fraxken 

Added this fix in the wrong commit bu it's there

fraxken (approved):

PierreDemailly (approved):

fraxken:
@allcontributors please add @jean-michelet for code, test, doc

PR #218, opened 2024-01-29
Create ts-source-parser package

jean-michelet:
This PR follows on from discussion [#214](https://github.com/NodeSecure/js-x-ray/issues/214), which proposes a new package in workspaces for parsing TypeScript files. 

## Usage example
```js
import { TsSourceParser } from "@nodesecure/ts-source-parser";

const parser = new TsSourceParser();
const body = parser.parse("const x: number = 5;");

console.log(body);
```

## Usage with `js-x-ray` 
```js
import { runASTAnalysis } from "@nodesecure/js-x-ray";
import { readFileSync } from "node:fs";

const { warnings, dependencies } = runASTAnalysis(
    readFileSync("./file.ts", "utf-8"),
    { customParser: new TsSourceParser() }
);

console.log(dependencies);
console.dir(warnings, { depth: null });
```


.github/workflows/ts-source-parser.yml
```diff
@@ -0,0 +1,41 @@
+# This workflow will do a clean installation of node dependencies, cache/restore them, build the source code and run tests across different versions of node
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-nodejs
+
+name: Node.js CI
```
PierreDemailly:
```suggestion
name: TsSourceParser CI
```

The idea is to have a different name for this workspace so we can identify it in runned jobs

.github/workflows/ts-source-parser.yml
```diff
@@ -0,0 +1,41 @@
...
+      with:
+        egress-policy: audit # TODO: change to 'egress-policy: block' after couple of runs
+
+    - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@b39b52d1213e96004bfcb1c61a8a6fa8ab84f3e8 # v4.0.1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm install
+    - run: npm test
```
PierreDemailly:
```suggestion
    - run: npm run test --workspace=workspaces/ts-source-parser
```

workspaces/ts-source-parser/package.json
```diff
@@ -0,0 +1,37 @@
+{
+    "name": "@nodesecure/ts-source-parser",
```
PierreDemailly:
```suggestion
  "name": "@nodesecure/ts-source-parser",
```
We indents with 2 spaces (need to update the whole file). If you use VSC do you have EditorConfig extension enabled ?

PierreDemailly (commented):
You also need to add the workspace in the root package.json
https://github.com/NodeSecure/js-x-ray/blob/f4b47c72b7996261df264f91757c33bfcac97406/package.json#L21

Everything else lgtm

.github/workflows/ts-source-parser.yml
jean-michelet: Ok, but this flag does not exist in other workspaces, do you want to add them as well?

PierreDemailly: Well, yes you can ^^

PierreDemailly (approved):

Issue #223, opened 2024-01-30
Use JsSourceParser as default parser for AstAnalyser class


fraxken:
The parser argument in the constructor of AstAnalyser class is currently required:

https://github.com/NodeSecure/js-x-ray/blob/master/src/AstAnalyser.js#L18

For me we should make it optional and assign `new JsSourceParser()` by default.

madina0801:
Hi, I would like to work on it! Could you please assign it to me?

jean-michelet:
Go for it Madina, I think the issue is quite simple.

PR #227, opened 2024-01-31
Use JsSourceParser as default parser for AstAnalyser class

madina0801:
Issue #223 
It was suspiciously easy, let me know if there are any issues :)

### DIFF
--- a/test/AstAnalyser.spec.js
+++ b/test/AstAnalyser.spec.js
@@ -229,7 +229,7 @@ describe("AstAnalyser", (t) => {
   describe("constructor", () => {
     it("should not throw an error when instantiated without a custom parser", () => {
       assert.doesNotThrow(() => {
-        const analyser = getAnalyser();
+        const analyser = new AstAnalyser();
         // perform basic operations
         const result = analyser.analyse("const foo = 'bar';");
         // compare array of keys to an empty array to ensure there are no dependencies in result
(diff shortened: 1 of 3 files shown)


PierreDemailly:
Thanks for contributing, could you update the type here? https://github.com/NodeSecure/js-x-ray/blob/7422acba002fc75059a016c6e10a29ec575a6437/types/api.d.ts#L81

fraxken:
You can probably write a new test in https://github.com/NodeSecure/js-x-ray/blob/master/test/AstAnalyser.spec.js

If we declare the class without any parser and then use it (it should not throw any error).

test/AstAnalyser.spec.js
```diff
@@ -225,6 +225,18 @@ describe("AstAnalyser", (t) => {
       assert.strictEqual(preparedSource, "\nconst yo = 'foo'\n");
     });
   });
+
+  describe("constructor", () => {
+    it("should not throw an error when instantiated without a custom parser", () => {
+      assert.doesNotThrow(() => {
+        const analyser = getAnalyser();
```
PierreDemailly: Actually this test is wrong.
The `getAmalyser()` function returns the `AstAnalyser` instanciated with `JsSourceParser()`.

```js
analyser = new AstAnalyser(new JsSourceParser());
```

```suggestion
        const analyser = new AstAnalyser();
```

PierreDemailly (changes requested):

PierreDemailly (approved):

fraxken (approved):

fraxken:
@allcontributors please add @madina0801 for code

Issue #221, opened 2024-01-30
Inject custom probes in AstAnalyser class


fraxken:
The goal of the task is to allow to package users (developers) to inject a new custom probe using the AstAnalyser

```js
const { AstAnalyser, JsSourceParser } = require("@nodesecure/js-x-ray");

new AstAnalyser({
  parser: new JsSourceParser(),
  probes: [
    // Any valid probe object here
  }
});
```

We probably need to instanciate the ProbeRunner class outside of SourceFile

PR #250, opened 2024-03-09
feat(customProbes): inject custom probes as param for AstAnalyser

tchapacan:
fix https://github.com/NodeSecure/js-x-ray/issues/221

### Inject custom probes as param for AstAnalyser

```js
const { AstAnalyser, JsSourceParser } = require("@nodesecure/js-x-ray");

const analyser = new AstAnalyser({
  parser: new JsSourceParser(),
  customProbes,
  skipDefaultProbes: true
});

```

> Allow developers to inject new custom probe using the AstAnalyser

Purpose of this PR is to propose a fix/feat for the issue https://github.com/NodeSecure/js-x-ray/issues/221 by modifying the constructor of some of the relevant class such as (ProbeRunner, ASTAnalyzer, SourceFile) : 

- [x] code
- [x] tests 
- [x] docs

I tried to keep it as simple as possible without modifying that much what was already done (i haven't change the ProbeRunner instanciation from SourceFile), but I'm open to suggestion. 

Don't hesitate to give any feedback/suggestion/proposition to improve this PR 

Result after (yes this probe below is perfectible ahah) : 

```js
const kIncriminedCodeSample = "const danger = 'danger';";

const customProbes = [
  {
    name: "customProbeUnsafeDanger",
    validateNode: (node, sourceFile) => [true]
    ,
    main: (node, options) => {
      const { sourceFile, data: calleeName } = options;
      if (node.declarations[0].init.value === "danger") {
        sourceFile.addWarning("unsafe-danger", calleeName, node.loc);
  
        return ProbeSignals.Skip;
      }
  
      return null;
    }
  }
];

const analyser = new AstAnalyser({
  parser: new JsSourceParser(),
  customProbes,
  skipDefaultProbes: true
});

const result = analyser.analyse(kIncriminedCodeSample);
console.log(result);
```

```
➜  js-x-ray git:(fix/221) ✗ node example.js
{
  idsLengthAvg: 0,
  stringScore: 0,
  warnings: [ { kind: 'unsafe-danger', location: [Array], source: 'JS-X-Ray' } ],
  dependencies: Map(0) {},
  isOneLineRequire: false
}
```


src/SourceFile.js
```diff
@@ -20,14 +20,20 @@ export class SourceFile {
...
+  constructor(sourceCodeString, customProbes = []) {
     this.tracer = new VariableTracer()
       .enableDefaultTracing()
       .trace("crypto.createHash", {
         followConsecutiveAssignment: true, moduleName: "crypto"
       });
 
-    this.probesRunner = new ProbeRunner(this);
+    if (Array.isArray(customProbes) && customProbes.length > 0) {
+      this.probesRunner = new ProbeRunner(this, customProbes);
```
fraxken:
```suggestion
      this.probesRunner = new ProbeRunner(this, [...ProbeRunner.Defaults, ...customProbes]);
```

fraxken: We need to keep default probes (maybe we could introduce an option to choose to keep them or not). But for me by default we extend probes (and not overwriting them)

tchapacan: agreed this would give more flexibility/control, and help passing the tests easier as well (when we inject one probe). 
What do you have in mind for that option ?
- append (default) : add custom probes to the list 
- replace : replace list of default probe by the custom one

fraxken: Yep that how I see it right now. Outside of that there is probably long term subject like "do we want to inject the probe at the beginning, at the end, or before a given probe ?". But I guess we can deal with that later :)

src/AstAnalyser.js
```diff
@@ -16,8 +16,10 @@ export class AstAnalyser {
    * @constructor
    * @param { SourceParser } [parser]
    */
-  constructor(parser = new JsSourceParser()) {
+  constructor(parser = new JsSourceParser(), customProbes = [], mergeMode = "append") {
```
jean-michelet: If we keep adding arguments, maybe we should pass an options object once and for all...
`mergeMode` should be a boolean if there is only two possibilities, e.g. `isReplacing = false`.

Also, remember to update the types here: https://github.com/NodeSecure/js-x-ray/blob/929c275468f28638eec452bfd01a1331ad061866/types/api.d.ts#L81

src/SourceFile.js
```diff
@@ -20,14 +20,22 @@ export class SourceFile {
   encodedLiterals = new Map();
   warnings = [];
 
-  constructor(sourceCodeString) {
+  constructor(sourceCodeString, customProbes = [], mergeMode = "append") {
```
jean-michelet: Same logic here, I think it's highly probable that new options will be added in the future.

src/SourceFile.js
```diff
@@ -20,14 +20,22 @@ export class SourceFile {
...
     this.tracer = new VariableTracer()
       .enableDefaultTracing()
       .trace("crypto.createHash", {
         followConsecutiveAssignment: true, moduleName: "crypto"
       });
 
-    this.probesRunner = new ProbeRunner(this);
+    let mergedProbes;
+    if (Array.isArray(customProbes) && customProbes.length > 0) {
```
jean-michelet: Shouldn't `customProbes` always be an array?
If we feel the need to enforce typing, I think we should throw a `TypeError` if our apis are misused.

jean-michelet: This is shorter like this:
 ```js
let probes = ProbeRunner.Defaults;
if (Array.isArray(customProbes) && customProbes.length > 0) {
  probes = mergeMode === "replace" ? customProbes : [...probes, ...customProbes];
}
```

File: test/utils/index.js [hunk 5]
```diff
@@ -38,7 +38,8 @@ export function getSastAnalysis(
       return this.sourceFile.dependencies;
     },
     execute(body) {
-      const probeRunner = new ProbeRunner(this.sourceFile, [probe]);
+      const probes = Array.isArray(probe) ? probe : [probe];
```
jean-michelet: When does `getSastAnalysis` get an array of probes?

jean-michelet (commented):
Just a few questions, congratulations on the work accomplished.

test/utils/index.js
tchapacan: i have removed this unecessary part, it was me playing arround with the code while initially debugging the issue, thanks

src/SourceFile.js
tchapacan: I have created an `astOptions` object containing `customProbes`and `isReplacing` : 

```
astOptions: {
  customProbes:[],
  isReplacing: false
}
```

It can be extended later if needed

src/AstAnalyser.js
tchapacan: same comment as here https://github.com/NodeSecure/js-x-ray/pull/250#discussion_r1520155954 lmk if i'm on the right track

I took a stab trying to start playing with the types, lmk if it's good

src/AstAnalyser.js
```diff
@@ -14,10 +14,12 @@ import { JsSourceParser } from "./JsSourceParser.js";
 export class AstAnalyser {
   /**
    * @constructor
-   * @param { SourceParser } [parser]
+   * @param {SourceParser} [parser]
+   * @param astOptions
    */
-  constructor(parser = new JsSourceParser()) {
+  constructor(parser = new JsSourceParser(), astOptions = { isReplacing: false, customProbe: [] }) {
```
jean-michelet: 1. You can define everything within an `options` object, including `customParser`. This approach won't cause any issues because this class was specifically designed for the upcoming major release, and the code is not currently available on npm:  https://www.npmjs.com/package/@nodesecure/js-x-ray?activeTab=code

2. By the way, the name `astOptions` is overly specific and misleading, as it suggests options for the abstract syntax tree (AST) rather than the AST analyzer itself. Simply using `options` is clear enough.

3. In your constructor, you have no guarantee that the user will not pass an empty `options` object or one with missing properties, which could lead to subtle bugs. It's better to treat the options passed to the constructor as a means to override default options rather than as a reliable source. For inspiration on handling default options, take a look at this example: https://github.com/NodeSecure/js-x-ray/blob/929c275468f28638eec452bfd01a1331ad061866/src/NodeCounter.js#L29

types/api.d.ts
```diff
@@ -47,10 +59,26 @@ interface RuntimeOptions {
    * @default false
    */
   removeHTMLComments?: boolean;
-  
+
   customParser?: SourceParser;
 }
 
+interface ASTOptions {
```
jean-michelet: The terminology is misleading, these are not options for the Abstract Syntax Tree ([AST](https://en.wikipedia.org/wiki/Abstract_syntax_tree)). Essentially, an AST is a high-level representation of the source code's syntax, which we traverse in order to identify security vulnerabilities.

tchapacan: I have tried to rework a little the type for the options (and the naming), not sure it's 100% correct but lmk if its better like this

src/AstAnalyser.js
tchapacan: i have put the options in an object finally as you recommended (including customProbes, customParser etc..) and renamed the astOptions to options. I have started to add some assert check in the ASTAnalyzer constructor as well to throw error if input config parameter are not clean, lmk if you have suggestion

index.js
```diff
@@ -9,10 +9,12 @@ function runASTAnalysis(
 ) {
   const {
     customParser = new JsSourceParser(),
+    customProbe = [],
+    isReplacing = false,
```
jean-michelet:
```suggestion
    replaceProbes = false,
```

index.js
```diff
@@ -9,10 +9,12 @@ function runASTAnalysis(
 ) {
   const {
     customParser = new JsSourceParser(),
+    customProbe = [],
```
jean-michelet:
```suggestion
    customProbes = [],
```

src/AstAnalyser.js
```diff
@@ -14,10 +15,25 @@ import { JsSourceParser } from "./JsSourceParser.js";
 export class AstAnalyser {
   /**
    * @constructor
-   * @param { SourceParser } [parser]
+   * @param options
```
jean-michelet:
```suggestion
   * @param {object} [options={}] 
```

You can also add the options params:
```js
@param {Object} [options={}] 

@param {Array} [options.customProbe=[]]
// etc.
```

src/AstAnalyser.js
```diff
@@ -14,10 +15,25 @@ import { JsSourceParser } from "./JsSourceParser.js";
...
+    }
+    if (options.isReplacing !== undefined) {
+      assert(typeof options.isReplacing === "boolean", `isReplacing must be a boolean`);
+    }
+
+    this.parser = options.customParser || new JsSourceParser();
+    this.options = {
+      isReplacing: options.isReplacing || false,
+      customProbe: options.customProbe || []
+    };
```
jean-michelet, line 27: I think we should keep lines 32 to 36 and delete the type assertions; They make the code too messy imo and we're already making the effort to offer relevant typing through annotations as well as TS types.

Also, use `??` instead of `||`, as they differ in their treatment of `falsy` and `nullish` values. 

Regarding the `options` property, it might be clearer to avoid creating it because it can be misleading. Instead, they should be handled as properties like `parser`. Eventually, we could introduce something like `options.probesOptions`, but I am not sure.

types/api.d.ts
```diff
@@ -34,21 +34,49 @@ interface Dependency {
   location?: null | SourceLocation;
 }
 
-interface RuntimeOptions {
+interface RuntimeCommonOptions {
```
jean-michelet: I think we're having difficulties maintaining relevant typing since `runASTAnalysis` and `runASTAnalysisOnFile` take arguments for `AstAnalyser` construction as well as `AstAnalyser.analyze` execution. Maybe, we should consider removing these functions in the next major and enforce the use of `AstAnalyser`.

index.js
PierreDemailly: Yeah `isReplacing` is not accurate.
Or `skipDefaultProbes`, falsy by default.
`keepDefaultProbes`, truthy by default, would also work.

src/AstAnalyser.js
```diff
@@ -14,10 +15,25 @@ import { JsSourceParser } from "./JsSourceParser.js";
...
    * @constructor
-   * @param { SourceParser } [parser]
+   * @param options
    */
-  constructor(parser = new JsSourceParser()) {
-    this.parser = parser;
+  constructor(options = {}) {
+    if (options.customParser !== undefined) {
+      assert(options.customParser instanceof JsSourceParser || typeof options.customParser === "object",
+        `customParser must be an instance of JsSourceParser or an object`);
```
PierreDemailly:
```suggestion
        "customParser must be an instance of JsSourceParser or an object");
```

src/AstAnalyser.js
```diff
@@ -14,10 +15,25 @@ import { JsSourceParser } from "./JsSourceParser.js";
...
+    }
+    if (options.customProbe !== undefined) {
+      assert(Array.isArray(options.customProbe), `customProbe must be an array`);
+    }
+    if (options.isReplacing !== undefined) {
+      assert(typeof options.isReplacing === "boolean", `isReplacing must be a boolean`);
```
PierreDemailly: I'm not sure for these `assert` because it throws `AssertionError`. Usually we wanna throw `TypeError` instead.
But its probably ok :)

jean-michelet: I still think we should get rid of these type tests, wdyt?
Checking for `undefined` value in order to set default values is enough if we don't use TS.

PierreDemailly: Yep, it should be ok

index.js
tchapacan: as proposed by @PierreDemailly i went for `skipDefaultProbes` LMK if thats ok

index.js
tchapacan: thanks its fixed

src/AstAnalyser.js
tchapacan: Further params has been added, thanks for spotting it !

src/AstAnalyser.js
tchapacan: The assert part has been removed, typing and default value are enough agreeed

src/AstAnalyser.js
tchapacan: according to this com https://github.com/NodeSecure/js-x-ray/pull/250#discussion_r1526343856 this part has been removed

types/api.d.ts
tchapacan: This should be part of this PR or an another one? As it is the entrypoint of the tools, how do you see it evolve without those 2 methods (link between other nodesecure component i might ignore) ? 
I agree that something have to be done, as it drive us to have a hacky way of representing the types between the "exec options" and "probe options" which can lead to confusion.

src/AstAnalyser.js
tchapacan: Thanks for `??` it's fixed in latest version (the assert has been removed as well to rely only on types)

i have rename the `options` property to `probeOptions` for better clarity. But maybe you were suggesting to change the structure of the options? Instead of remapping some parameters in object options while runtime between each constructor, directly consume a fix structure of options supplied initially by the user? Something like that (below) could be passed between necessary class and used accordingly ?

```js
{
  customParser: new JsSourceParser(),
  probesOptions: {
    customProbes: [],
    skipDefaultProbes: false
  },
  runtimeOptions: {
    minified: false,
    package: false,
  etc
  ...
  },
}
```

It might be link to this subject as well https://github.com/NodeSecure/js-x-ray/pull/250#discussion_r1525834884

types/api.d.ts
jean-michelet: I am not sure what should be done, it was a suggestion.
Are you okay to remove these methods @fraxken or what do you think should be done here?

fraxken: Right now that's not totally clear to me. From the general discussion I would say to deal with that in a later PR (no need to block this one).

src/AstAnalyser.js
```diff
@@ -1,6 +1,7 @@
 // Import Node.js Dependencies
 import fs from "node:fs/promises";
 import path from "node:path";
+import assert from "node:assert";
```
jean-michelet: Remove that line pls.
Don't forget to run `npm run lint` before push.

src/AstAnalyser.js
```diff
@@ -14,10 +15,17 @@ import { JsSourceParser } from "./JsSourceParser.js";
 export class AstAnalyser {
   /**
    * @constructor
-   * @param { SourceParser } [parser]
+   * @param {object} [options={}]
+   * @param {JsSourceParser} [options.customParser]
```
jean-michelet:
```suggestion
   * @param {SourceParser} [options.customParser]
```

test/AstAnalyser.spec.js
```diff
@@ -1,6 +1,6 @@
 // Import Node.js Dependencies
 import { describe, it } from "node:test";
-import assert from "node:assert";
+import assert, { AssertionError } from "node:assert";
```
jean-michelet: Same here.

test/issues/221-inject-custom-probes.spec.js
```diff
@@ -0,0 +1,90 @@
...
+
+test("should replace list of probes", () => {
+  const analyser = new AstAnalyser({
+    parser: new JsSourceParser(),
+    customProbes,
+    skipDefaultProbes: true
+  });
+  const result = analyser.analyse(kIncriminedCodeSample);
+
+  console.log(analyser);
```
jean-michelet:
```suggestion
```

jean-michelet (commented):
After that, should be ok :+1:

test/issues/221-inject-custom-probes.spec.js
PierreDemailly, line 1: I don't think the file need to be in `/test/issues`
The `issues` folder is reserved to bugs / edge cases (otherwise almost every test would be in this folder 😅)

Wait for @fraxken to confirm

PierreDemailly:
https://github.com/NodeSecure/js-x-ray/blob/master/index.js
We need to add theses new options to `runASTAnalysis` & `runASTAnalysis`

test/issues/221-inject-custom-probes.spec.js
fraxken: Yes we should include those tests in a proper file (and not in issues)

src/AstAnalyser.js
tchapacan: resolving as updated in latest commit

test/issues/221-inject-custom-probes.spec.js
tchapacan: i have updated in latest commit by trying to move thoses tests in the relevant tests files

i still have to update readme

test/AstAnalyser.spec.js
```diff
@@ -236,6 +265,13 @@ describe("AstAnalyser", (t) => {
         assert.deepEqual([...result.dependencies.keys()], []);
       });
     });
+
+    it("should instantiate with correct default ASTOptions", () => {
+      const analyser = new AstAnalyser();
+      assert(analyser.parser instanceof JsSourceParser || typeof analyser.parser.customParser === "object");
```
jean-michelet:
```suggestion
      assert(analyser.parser instanceof JsSourceParser);
```

jean-michelet (commented):
Lgtm, with a ~~nit~~

test/AstAnalyser.spec.js
jean-michelet: Not sure if you should use alias `assert.ok`, it could be more in line with the overall style in the codebase.

tchapacan: Ok for the assert.ok i added that condition to that test due to the fakeparser test if i remember correctly, gonna have a look and will push later, thanks again for the review 💪

jean-michelet: I don't understand `analyser.parser.customParser`.
Anyway, we test default parser here, not a custom one (`FakeSourceParser`).

jean-michelet: Btw:
should instantiate with correct default ~~ASTOptions~~ options.

tchapacan: thanks it should be fixed now in latest version

tchapacan:
I have proposed a quick update of the readme, if you think it needs more details i can try create a dedicated .md file for the customProbe part ? and refer it in the readme briefly  ?

README.md
```diff
@@ -134,15 +134,88 @@ This section describe all the possible warnings returned by JSXRay. Click on the
...
+
+A probe is a pair of two functions (`validateNode` and `main`) that will be called on each node of the AST. It will return a warning if the pattern is detected.
+Below a basic probe that detect a string assignation to `danger`:
+
+```ts
+export const customProbes = [
+  {
+    name: "customProbeUnsafeDanger",
+    validateNode: (node, sourceFile) => [node.type === "VariableDeclaration" && node.declarations[0].init.value === "danger"]
+    ,
```
PierreDemailly:
```suggestion
    validateNode: (node, sourceFile) => [
      node.type === "VariableDeclaration" && node.declarations[0].init.value === "danger"
    ],
```

The goal is to prevent x-scroll in the code snippet

README.md
```diff
@@ -134,15 +134,88 @@ This section describe all the possible warnings returned by JSXRay. Click on the
...
+  warnings: [ { kind: 'unsafe-danger', location: [Array], source: 'JS-X-Ray' } ],
+  dependencies: Map(0) {},
+  isOneLineRequire: false
+}
+```
+
+Congrats, you have created your first custom probe! 🎉
+
+> Check the types in [index.d.ts](index.d.ts) and [types/api.d.ts](types/api.d.ts) for more details about the `options`
+>
```
PierreDemailly:
```suggestion
> [!TIP]
> Check the types in [index.d.ts](index.d.ts) and [types/api.d.ts](types/api.d.ts) for more details about the `options`
```

README.md
```diff
@@ -134,15 +134,88 @@ This section describe all the possible warnings returned by JSXRay. Click on the
...
+```
+
+Congrats, you have created your first custom probe! 🎉
+
+> Check the types in [index.d.ts](index.d.ts) and [types/api.d.ts](types/api.d.ts) for more details about the `options`
+> 
 ## API
 <details>
-<summary>runASTAnalysis(str: string, options?: RuntimeOptions): Report</summary>
+<summary>declare function runASTAnalysis(str: string, options?: RuntimeOptions & AstAnalyserOptions): Report</summary>
```
PierreDemailly:
```suggestion
<summary>runASTAnalysis(str: string, options?: RuntimeOptions & AstAnalyserOptions): Report</summary>
```
I'm not sure we need to add `declare function`, we document functions wihout it in all our projects
(same below)

PierreDemailly:
> I have proposed a quick update of the readme, if you think it needs more details i can try create a dedicated .md file for the customProbe part ? and refer it in the readme briefly ?

The update is fine 👍 I don't think we need a dedicated doc for the moment

tchapacan:
> > I have proposed a quick update of the readme, if you think it needs more details i can try create a dedicated .md file for the customProbe part ? and refer it in the readme briefly ?
> 
> The update is fine 👍 I don't think we need a dedicated doc for the moment

Thanks for the review, I have fix the little typo in the readme in latest commit, lmk if it seems good enough

PierreDemailly (approved):
LGTM

fraxken (approved):

PR #253, opened 2024-03-19
[WIP] Rework sourcefile analysis strategy

### LINKED ISSUE #237: Rework SourceFile analysis strategy
Right now JS-X-Ray is only capable to scan one sourcefile by one sourcefile. The [Scanner](https://github.com/NodeSecure/scanner) package is currently responsible of listing and iterating all JavaScript files from a given NPM tarball (for example).

What's the issue with that?

```
index.js
src/
  other.js
test/
  foobar.js
```

In the example above we will scan every files. But in reality there is a high probability that `test/foobar.js` will never be executed (and it will also be the biggest vector of false positives).

My idea is to add a new strategy that will take **entry files** as input. We will then only scan files imported from these entry points.

Eventually, we could combine the two ways of doing things to ensure greater security while reducing false positives overall.

jean-michelet:
Fixes #237 

It's still a WIP.
I don't know if this is what you were thinking @fraxken ....

I'd like to allow the use of blobs: 
```js
["src/**/*", "tests/**/*"]
```

I'll adapt the code when [Tchapacan PR](https://github.com/NodeSecure/js-x-ray/pull/250) is merged.

### DIFF
--- a/src/AstAnalyser.js
+++ b/src/AstAnalyser.js
@@ -16,9 +16,10 @@ export class AstAnalyser {
    * @constructor
    * @param { SourceParser } [parser]
    */
-  constructor(parser = new JsSourceParser()) {
+  constructor(parser = new JsSourceParser(), entryFiles = []) {
     this.parser = parser;
     this.analyzedFiles = new Map();
+    this.entryFiles = entryFiles.map((entry) => path.resolve(entry));
   }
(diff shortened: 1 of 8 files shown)


jean-michelet:
I will reopen a clean one when I'm done.

PR #258, opened 2024-04-12
Create EntryFilesAnalyzer class to analyze a set of entry files

### LINKED ISSUE #237: Rework SourceFile analysis strategy

jean-michelet:
Fixes #237


jean-michelet:
Oops, I forgot to add the types, I'll do this.

test/EntryFilesAnalyser.spec.js
```diff
@@ -0,0 +1,95 @@
...
+    const generator = entryFilesAnalyser.analyse([entryUrl]);
+
+    await assertReport(generator, entryUrl, true);
+    await assertReport(generator, new URL("deps/default.js", FIXTURE_URL), true);
+    await assertReport(generator, new URL("deps/default.cjs", FIXTURE_URL), true);
+    await assertReport(generator, new URL("deps/default.mjs", FIXTURE_URL), true);
+    await assertReport(generator, new URL("deps/default.node", FIXTURE_URL), true);
+    await assertReport(generator, new URL("deps/default.jsx", FIXTURE_URL), true);
+
+    await assertAllReportsYielded(generator);
```
fraxken, line 70: Probably better do to something like that

```
const result = await Array.fromAsync(generator);

// assert result
```

fraxken: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fromAsync

jean-michelet: Because not part of Node LTS yet, what do you suggest?

fraxken: For now that ok

jean-michelet:
Any suggestion @PierreDemailly?

fraxken (approved):

test/fixtures/entryFiles/deps/default.mjs
```diff
@@ -0,0 +1 @@
+import internalDep from 'internalDep';
```
PierreDemailly: As internal imports must be relative (with extension) in ESM, i'm not sure to understand here.
Should it be `import externalDep from 'externalDep';` ?
Should it be `import internalDep from './internalDep.js';` + need to create `internalDep.js` file ?
If i'm not wrong, there are no ESM-import tests, should we add some ?

Same question in `default.js` -> `require('internalDep')` what is intended ?

PierreDemailly:
Should we handle dynamic imports ? (In this PR or in a future one)

PierreDemailly (approved):

jean-michelet:
> Should we handle dynamic imports ? (In this PR or in a future one)

#259

fraxken (approved):

PR #269, opened 2024-05-15
feat: add intialize and finalize for AstAnalyser.analyze API

### LINKED ISSUE #268: [AstAnalyser] Add initialize and finalize callback to manipulate SourceFile
Current AstAnalyser as flaw when we want to manipulate the SourceFile class before or after walking the AST, here is an example where we add new custom tracing;

```js
import { AstAnalyser } from "@nodesecure/js-x-ray";

const scanner = new AstAnalyser();

scanner.analyse("const foo = 'bar';", {
 initialize(sourceFile) {
   sourceFile.tracer.trace("...");
 },
 finalize(sourceFile) {
   // do something
 }
});
```

Those callbacks should be triggered before and after walking the AST

image

tony-go:
Fixes #268

Signed-off-by: Tony Gorez <gorez.tony@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,6 +223,8 @@ interface RuntimeOptions {
   module?: boolean;
   removeHTMLComments?: boolean;
   isMinified?: boolean;
+  initialize?: (sourceFile: SourceFile) => void;
+  finalize?: (sourceFile: SourceFile) => void;
 }
 ```
 
--- a/docs/api/AstAnalyser.md
+++ b/docs/api/AstAnalyser.md
@@ -64,3 +64,29 @@ type ReportOnFile = {
   warnings: Warning[];
 }
 ```
+
+## Examples
+
+### `initialize`/`finalize` Hooks
+
+The `analyse` method allows for the integration of two hooks: `initialize` and `finalize`. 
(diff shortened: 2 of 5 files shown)


tony-go:
All tests should failed, I'd like to confirm that we are okay with the API first.

README.md
```diff
@@ -223,6 +223,8 @@ interface RuntimeOptions {
   module?: boolean;
   removeHTMLComments?: boolean;
   isMinified?: boolean;
+  initialize?: (sourceFile: SourceFile) => void;
+  finalize?: (sourceFile: SourceFile) => void;
```
tony-go, line 227: More than documenting these types, I could add a proper example somewhere.

fraxken: probably somewhere in /docs/api/AstAnalyser.md

fraxken:
Note: we need the same for analyseFile (it use analyse but interfaces/options should be the same)

tony-go:
Hey @fraxken could you PTAL?

> Note: we need the same for analyseFile (it use analyse but interfaces/options should be the same)

Lets do it into a subsequent PR, if you don't mind.

PR #271, opened 2024-05-17
feat: enable initalize and finalize for AstAnalyser.analyseFile

tony-go:


test/AstAnalyser.spec.js
```diff
@@ -254,30 +254,115 @@ describe("AstAnalyser", (t) => {
...
+          });
 
-  it("should fail with a parsing error", async () => {
-    const result = await getAnalyser().analyseFile(
-      new URL("parsingError.js", FIXTURE_URL),
-      { module: false, packageName: "foobar" }
-    );
+          assert.strictEqual(res.ok, false);
+          assert.strictEqual(res.warnings[0].value, "options.initialize must be a function");
+          assert.strictEqual(res.warnings[0].kind, "parsing-error");
```
tony-go, line 297: @fraxken that doesn't make sense to me, but it seems that every error catched by `analyseFile` are wrapped into this result object.

So either we let it like that, or I throw like for the `analyze` function, but we'll do the check twice in this case (as the `analyse` func will do it).

fraxken (approved):

PR #320, opened 2025-05-26
feat(EntryFilesAnalyzer): add option to ignore entry file that doesn't exist

### LINKED ISSUE #319: EntryFilesAnalyser: add option to ignore entry file that doesn't exist
Current API eg `analyse` throw if provided with entry files that don't exist, the idea would be to provide a new option to the constructor to ignore the entry if they don't exist.

```js
const entries = ["index.js"]; // this file doesn't exist
const efa = new EntryFilesAnalyser({
  ignoreENOENT: true
});

for await (const fileReport of efa.analyse(entries)) {
}

// No error throw
```

I guess we can use DiGraph and `dependencies` property to tell the number of provided entries doesn't match with the number of vertices.

clemgbld:
fix #319


fraxken:
@allcontributors please add @clemgbld for code,test

PR #334, opened 2025-06-02
test(AstAnalyser): add tests for flags property and minified/oneline-require detection

intincrab:
### Tests

1. **`should include flags property in response`** - Ensures flags property exists and prevents regression where it was forgotten
2. **`should add is-minified flag for minified files`** - Tests minified content detection and flag setting  
3. **`should add oneline-require flag for one-line exports`** - Tests one-line require exports and flag mutual exclusion

#### Coverage
- Flags property existence
- "is-minified" flag functionality  
- "oneline-require" flag functionality
- Flag mutual exclusion logic

### DIFF
--- a/test/AstAnalyser.spec.js
+++ b/test/AstAnalyser.spec.js
@@ -2,7 +2,7 @@
 // Import Node.js Dependencies
 import { describe, it } from "node:test";
 import assert from "node:assert";
-import { readFileSync } from "node:fs";
+import { readFileSync, writeFileSync, unlinkSync } from "node:fs";
 
 // Import Internal Dependencies
 import { AstAnalyser, JsSourceParser } from "../index.js";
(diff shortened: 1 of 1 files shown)


test/AstAnalyser.spec.js
```diff
@@ -458,6 +460,53 @@ describe("AstAnalyser", (t) => {
...
+      const result = getAnalyser().analyseFileSync(
+        new URL("depName.js", FIXTURE_URL)
+      );
+
+      assert.ok(result.ok);
+      assert.ok(result.flags instanceof Set);
+    });
+
+    it("should add is-minified flag for minified files", () => {
+      const minifiedContent = "var a=require(\"fs\"),b=require(\"http\");" +
```
PierreDemailly:
```suggestion
    it("should add is-minified flag for minified files", (t) => {
      t.plan(3);
      const minifiedContent = "var a=require(\"fs\"),b=require(\"http\");" +
```
Because the assertions are inside a `try-finally` block, if for some reason `getAnalyser().analyseFileSync(tempMinFile);` throws, the tests will still pass.
`t.plan(3)` prevents this.

(same for others tests)

intincrab: Tried adding `t.plan(3)`, but `npm test` fails. I might be missing something?

PierreDemailly: Oh yeah, my bad.
You also need to replace `assert` with `t.assert`.

> To make sure assertions are tracked, t.assert must be used instead of assert directly.

intincrab: this worked! thanks

test/AstAnalyser.spec.js
```diff
@@ -460,16 +460,18 @@ describe("AstAnalyser", (t) => {
       assert.strictEqual(parsingError.kind, "parsing-error");
     });
 
-    it("should include flags property in response", () => {
+    it("should include flags property in response", (t) => {
+      t.plan(2);
```
PierreDemailly: it was not necessary here ^^
The assertions in this test are not in a `try-catch/finally` block, so if `getAnalyser().analyseFileSync` throws, test will already fail

intincrab: yes, my bad. i've fixed it now.

PierreDemailly (approved):

fraxken (approved):

PR #351, opened 2025-06-24
refactor: migrate estree-ast-utils to TypeScript

fraxken:


PierreDemailly (approved):

workspaces/estree-ast-utils/src/extractLogicalExpression.ts
```diff
@@ -1,7 +1,6 @@
-
 export function* extractLogicalExpression(
-  node
-) {
+  node: any
+): IterableIterator<{ operator: "||" | "&&" | "??", node: any }> {
```
tony-go: There is no way ti put that into a typedef `"||" | "&&" | "??"` ?

workspaces/estree-ast-utils/test/utils.ts
```diff
@@ -17,10 +17,14 @@ export function codeToAst(code) {
   return estreeRootNode.body;
 }
 
-export function getExpressionFromStatement(node) {
+export function getExpressionFromStatement(node: any) {
```
tony-go, line 20: As `node` could be complex to type I wonder if we could not create a Node type which points to `any` until we found better.

tony-go (approved):

PR #352, opened 2025-06-25
refactor: migrate tracer to TypeScript

fraxken:


workspaces/tracer/src/utils/notNullOrUndefined.ts
```diff
@@ -0,0 +1,5 @@
+export function notNullOrUndefined(
+  value: any
+): value is NonNullable<any> {
+  return value !== null && value !== void 0;
```
tony-go, line 4: JS question (because I did not remember): the `value !== void 0` is the same as `value !== undefined` ?

fraxken: Yes

workspaces/tracer/package.json
```diff
@@ -25,6 +26,7 @@
   },
   "homepage": "https://github.com/NodeSecure/js-x-ray/tree/master/workspaces/tracer#readme",
   "dependencies": {
-    "@nodesecure/estree-ast-utils": "^3.0.0"
+    "@nodesecure/estree-ast-utils": "^3.0.0",
+    "meriyah": "^6.1.3"
```
PierreDemailly, line 30: `meriyah` should be dev dep since its used only for typing ?

fraxken: You can't, how types work if that's devDep ? :p 

It's not a big deal since that's anyway a dep of JS-X-Ray

PierreDemailly (approved):

PR #355, opened 2025-06-27
feat: new optional synchronous IO warning

### LINKED ISSUE #349: Optional Warning: usage of synchronous I/O
Detect usage of synchronous API from Node.js core such as fs or child_process. The idea is to implement a new kind of warning that are optional by default.

This new warning can be useful in the context of a API (usage of sync method that can freeze the event loop for example).

clemgbld:
Closes #349


README.md
```diff
@@ -106,7 +107,8 @@ type WarningName = "parsing-error"
 | "weak-crypto"
 | "unsafe-import"
 | "unsafe-command"
-| "shady-link";
+| "shady-link"
+| "perf";
```
PierreDemailly: I'm wondering if `perf` is too generic 🤔
`sync-io` would make more sense imo.

For instance when analyzing a CLI project, it makes more sense to disable `sync-io` than `perf`, because it's not relevant in that context: it's not a server API.
That said, you're still somewhat concerned with "performance" in a broader sense.

WDYT?

clemgbld: Yes if you look at the comment of @tony-go on the issue, we could extend the warning to detect other pattern than sync-io.
But we could also make new warnings for each specific patterns.

PierreDemailly: I prefer specific ones. This would mean that `unsafe-regex` could be dropped to be merged into `perf` if we go this way, i'm not +1 with this idea.
But let's see what others think :)

fraxken: Yep the warning should be `synchronous-io` (same for the probe etc)

clemgbld: Got it

src/probes/isSyncIO.js
```diff
@@ -0,0 +1,27 @@
+// Import Third-party Dependencies
+import { getCallExpressionIdentifier } from "@nodesecure/estree-ast-utils";
+
+// Constants
+const kModules = ["fs", "crypto", "child_process", "zlib"];
```
fraxken:
```suggestion
// CONSTANTS
const kTracedNodeCoreModules = ["fs", "crypto", "child_process", "zlib"];
```

src/probes/isSyncIO.js
```diff
@@ -0,0 +1,27 @@
...
+
+function validateNode(node, { tracer }) {
+  const id = getCallExpressionIdentifier(node, { tracer });
+  if (id === null || !kModules.some((m) => tracer.importedModules.has(m))) {
```
fraxken: Hmm is there is no lint warning here? I would rename `m` to `moduleName`

src/SourceFile.js
```diff
@@ -26,7 +26,115 @@ export class SourceFile {
...
+        moduleName: "zlib"
+      })
+      .trace("zlib.brotliCompressSync", {
+        followConsecutiveAssignment: true,
+        moduleName: "zlib"
+      })
+      .trace("zlib.brotliDecompressSync", {
+        followConsecutiveAssignment: true,
+        moduleName: "zlib"
       });
```
fraxken: I wonder if we can improve that doing something like 

```js
["fs.readFileSync"].forEach((identifierOrMemberExp) => ...
```

A then just `identifierOrMemberExp.split(".")[0]` for moduleName

src/probes/isSyncIO.js
clemgbld: nope i got no warning, got it

src/SourceFile.js
```diff
@@ -25,9 +55,12 @@ export class SourceFile {
   constructor(sourceCodeString, probesOptions = {}) {
     this.tracer = new VariableTracer()
       .enableDefaultTracing()
-      .trace("crypto.createHash", {
-        followConsecutiveAssignment: true, moduleName: "crypto"
-      });
+    
+kIdentifierOrMemberExps.forEach((identifierOrMemberExp) =>
```
fraxken: Not syntaxed properly

fraxken:
think to add new line at the end of each fixtures files

docs/synchronous-io.md
```diff
@@ -0,0 +1,19 @@
+# Synchronous IO
+
+| Code | Severity | i18n | Experimental |
+| --- | --- | --- | :-: |
+| synchronous-io | `Warning` | `sast_warnings.sync-io` | ✔️ |
```
fraxken:
```suggestion
| synchronous-io | `Warning` | `sast_warnings.synchronous-io` | ✔️ |
```

To avoid difference

src/warnings.js
```diff
@@ -56,6 +56,11 @@ export const warnings = Object.freeze({
     i18n: "sast_warnings.unsafe-command",
     severity: "Warning",
     experimental: true
+  },
+  "sync-io": {
+    i18n: "sast_warnings.sync-io",
+    severity: "Warning",
+    experimental: true
   }
```
fraxken:
```suggestion
  "synchronous-io": {
    i18n: "sast_warnings.synchronous-io",
    severity: "Warning",
    experimental: true
  }
```

src/SourceFile.js
```diff
@@ -24,15 +54,29 @@ export class SourceFile {
...
 
     let probes = ProbeRunner.Defaults;
     if (Array.isArray(probesOptions.customProbes) && probesOptions.customProbes.length > 0) {
       probes = probesOptions.skipDefaultProbes === true ? probesOptions.customProbes : [...probes, ...probesOptions.customProbes];
     }
+
+    if (Array.isArray(probesOptions.optionalWarnings)) {
+      const optionalProbes = probesOptions.optionalWarnings
+        .map((warning) => ProbeRunner.Optionals[warning])
+        .filter(Boolean);
```
fraxken:
```suggestion
        .flatMap((warning) => ProbeRunner.Optionals[warning] ?? []);
```

Probably work

fraxken (approved):

fraxken (approved):

Issue #353, opened 2025-06-26
Add new initialize and finalize callbacks for probes


fraxken:
# Problem

Currently, there is no way to execute probe-related code during the initialization or teardown phases of the [SourceFile](https://github.com/NodeSecure/js-x-ray/blob/master/workspaces/js-x-ray/src/SourceFile.ts) class lifecycle.

# Proposal

My idea is to introduce two new optional callbacks for probes, to be implemented within the ProbeRunner:

- [x] `initialize(SourceFile)` (done with PR: https://github.com/NodeSecure/js-x-ray/pull/362)
- [ ] `finalize(SourceFile)`

Both methods would receive the current SourceFile instance as a parameter. Since ProbeRunner already stores this reference internally, this integration would be straightforward.

> [!NOTE]
> SourceFile has no finalize/teardown lifecycle, we need to implement one and use in AstAnalyser (probably)

## Example

Here’s a snippet of code that would be moved into the initialize callback of the `isWeakCrypto` probe:

```js
this.tracer = new VariableTracer()
  .enableDefaultTracing()
  // ↓↓↓ this part
  .trace("crypto.createHash", {
    followConsecutiveAssignment: true, moduleName: "crypto"
  });
```

We don't need to trace `crypto.createHash` if the isWeakCrypto probe is not used.

![Image](https://github.com/user-attachments/assets/9e7dca6f-1881-423b-9fb1-ab8c55736a7a)

Tchoupinax:
The new link of the [SourceFile](https://github.com/NodeSecure/js-x-ray/blob/master/workspaces/js-x-ray/src/SourceFile.ts)

fraxken:
> The new link of the [SourceFile](https://github.com/NodeSecure/js-x-ray/blob/master/workspaces/js-x-ray/src/SourceFile.ts)

Fixed

clemgbld:
@fraxken What is the difference between the finalize that we want to implement and the teardown function used in isRequire ?

fraxken:
> [@fraxken](https://github.com/fraxken) What is the difference between the finalize that we want to implement and the teardown function used in isRequire ?

Teardown is the finally of the probe (so it will always execute even if there is an error).

![Image](https://github.com/user-attachments/assets/5cb37d31-5985-4aaa-85c7-d920073a899b)

Finalize is a bit more related to the life cycle of SourceFile and would execute at the very end of the anaysis.

PR #362, opened 2025-06-28
feat(probes): add initialize

clemgbld:
#353

### DIFF
--- a/workspaces/js-x-ray/src/probes/isSyncIO.js
+++ b/workspaces/js-x-ray/src/probes/isSyncIO.js
@@ -45,12 +45,10 @@ function validateNode(node, { tracer }) {
 }
 
 function initialize(sourceFile) {
-  if (sourceFile) {
-    kSyncIOIdentifierOrMemberExps.forEach((identifierOrMemberExp) => sourceFile.tracer.trace(identifierOrMemberExp, {
-      followConsecutiveAssignment: true,
-      moduleName: identifierOrMemberExp.split(".")[0]
-    }));
-  }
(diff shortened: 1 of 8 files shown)


workspaces/js-x-ray/src/probes/isSyncIO.js
```diff
@@ -15,6 +44,15 @@ function validateNode(node, { tracer }) {
   return [data !== null && data.identifierOrMemberExpr.endsWith("Sync")];
 }
 
+function initialize(sourceFile) {
+  if (sourceFile) {
```
fraxken: ? sourceFile should never be undefined

clemgbld: There are some existing ProbeRunner tests where sourceFile is null.
But I guess you are right this is not possible from the end user perspective

workspaces/js-x-ray/test/ProbeRunner.spec.js
```diff
@@ -61,7 +59,21 @@ describe("ProbeRunner", () => {
...
 
       function instantiateProbeRunner() {
-        return new ProbeRunner(null, [fakeProbe]);
+        return new ProbeRunner(new SourceFile(""), [fakeProbe]);
+      }
+
+      assert.throws(instantiateProbeRunner, Error, "Invalid probe");
+    });
+
+    it("should fail if initialize initialize is present and not a function", () => {
```
PierreDemailly:
```suggestion
    it("should fail if initialize is present and not a function", () => {
```

fraxken (approved):

Issue #359, opened 2025-06-27
serialize-environment warning


fraxken:
The goal of the task is to trace `process.env` and create a new probe to detect usage of `JSON.stringify(process.env)`

Inspiration from guarddog: https://github.com/DataDog/guarddog/blob/main/guarddog/analyzer/sourcecode/npm-serialize-environment.yml

PR #363, opened 2025-06-28
feat(probes): add serialize-environment warning detection

intincrab:
#359

### DIFF
--- a/workspaces/js-x-ray/src/probes/isSerializeEnv.js
+++ b/workspaces/js-x-ray/src/probes/isSerializeEnv.js
@@ -1,3 +1,6 @@
+// Import Third-party Dependencies
+import { getCallExpressionIdentifier, getMemberExpressionIdentifier } from "@nodesecure/estree-ast-utils";
+
 // Import Internal Dependencies
 import { ProbeSignals } from "../ProbeRunner.js";
 
@@ -10,15 +13,8 @@ import { ProbeSignals } from "../ProbeRunner.js";
  * JSON.stringify(process[`env`])
  */
 function validateNode(node) {
-  if (node.type !== "CallExpression") {
-    return [false];
-  }
-
-  if (
-    node.callee.type !== "MemberExpression" ||
-    node.callee.object.name !== "JSON" ||
-    node.callee.property.name !== "stringify"
-  ) {
+  const id = getCallExpressionIdentifier(node);
+  if (id !== "JSON.stringify") {
     return [false];
   }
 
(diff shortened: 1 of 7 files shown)


PierreDemailly:
Could you add documentation (`/docs/serialize-environment.md`) please ?

workspaces/js-x-ray/test/probes/isSerializeEnv.spec.js
```diff
@@ -0,0 +1,56 @@
...
+
+test("should not detect other JSON.stringify calls", () => {
+  const str = "JSON.stringify({ foo: 'bar' })";
+  const ast = parseScript(str);
+  const sastAnalysis = getSastAnalysis(str, isSerializeEnv).execute(ast.body);
+
+  assert.strictEqual(sastAnalysis.warnings().length, 0);
+});
+
+test("should not detect non-JSON.stringify calls", () => {
```
clemgbld, line 50: I wonder if we should add a test that check that edge case like that are handled:
```js
const env = process.env;
JSON.stringify(env);
```

workspaces/js-x-ray/src/probes/isSerializeEnv.js
```diff
@@ -0,0 +1,60 @@
...
+ * JSON.stringify(process["env"])
+ * JSON.stringify(process[`env`])
+ */
+function validateNode(node) {
+  if (node.type !== "CallExpression") {
+    return [false];
+  }
+
+  if (
+    node.callee.type !== "MemberExpression" ||
```
clemgbld, line 18: I guess you can use the `getMemberExpressionIdentifer` from `@nodesecure/estree-ast-utils`.
it will handle edge case such as:

```js
JSON["s" + "tringify"](process.env);
```

workspaces/js-x-ray/src/probes/isSerializeEnv.js
```diff
@@ -0,0 +1,60 @@
...
+    node.callee.object.name !== "JSON" ||
+    node.callee.property.name !== "stringify"
+  ) {
+    return [false];
+  }
```
fraxken: You can probably use utils `getCallExpressionIdentifier` here

workspaces/js-x-ray/src/probes/isSerializeEnv.js
```diff
@@ -0,0 +1,60 @@
...
+  if (
+    firstArg.type === "MemberExpression" &&
+    firstArg.object.name === "process" &&
+    (
+      // Check for process.env
+      (firstArg.property.type === "Identifier" && firstArg.property.name === "env") ||
+      // Check for process["env"] or process["env"] or process[`env`]
+      (firstArg.property.type === "Literal" && firstArg.property.value === "env")
+    )
+  ) {
```
fraxken: You can probably also use helper `getMemberExpressionIdentifier`

workspaces/js-x-ray/test/probes/isSerializeEnv.spec.js
fraxken: I guess we can do it in a second PR by adding initialize and trace what's required

fraxken:
Note: you also need to update README.md

fraxken (approved):

PR #382, opened 2025-07-05
feat(probes) add finalize callback

### LINKED ISSUE #353: Add new initialize and finalize callbacks for probes

clemgbld:
Closes #353

### DIFF
--- a/workspaces/js-x-ray/test/AstAnalyser.spec.ts
+++ b/workspaces/js-x-ray/test/AstAnalyser.spec.ts
@@ -355,7 +355,7 @@ describe("AstAnalyser", () => {
       assert.equal(result.warnings.length, 1);
     });
 
-    it("should call initialize and finialize of every probes at the end", async(t) => {
+    it("should call initialize and finalize of every probes at the end", async(t) => {
       const calls: string[] = [];
       await new AstAnalyser(
         {
--- a/workspaces/js-x-ray/test/AstAnalyser.spec.ts
+++ b/workspaces/js-x-ray/test/AstAnalyser.spec.ts
@@ -355,6 +355,33 @@ describe("AstAnalyser", () => {
       assert.equal(result.warnings.length, 1);
     });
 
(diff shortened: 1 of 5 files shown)


workspaces/js-x-ray/test/AstAnalyser.spec.ts
```diff
@@ -355,6 +355,33 @@ describe("AstAnalyser", () => {
       assert.equal(result.warnings.length, 1);
     });
 
+    it("should call initialize and finialize of every probes at the end", async(t) => {
```
PierreDemailly:
```suggestion
    it("should call initialize and finalize of every probes at the end", async(t) => {
```

PierreDemailly (approved):

fraxken (approved):

PR #388, opened 2025-07-09
refactor(ProbeRunner): use Symbol.dispose

fraxken:


clemgbld (approved):

tony-go:
I like the idea ;) 

Do we have tests to update to see gains?

fraxken:
> I like the idea ;)
> 
> Do we have tests to update to see gains?

That's mainly a syntax thing here, tests are already here and validating that in both cases teardown is rightly triggered.

Issue #279, opened 2024-06-26
os, http(s), dns -> exfiltration ?


fraxken:
A lot of malicious package extract informations using the same pattern (os, dns and http). I think the idea is to flag any file with a warning `data-exfiltration` when we detect those.

Some examples extracted from socket.dev:
- https://socket.dev/npm/package/execer/files/8.0.7/index.json
- https://socket.dev/npm/package/reabctjs/files/1.1.4/new.js
- https://socket.dev/npm/package/all-star-2019/files/14.8.3/index.js
- https://socket.dev/npm/package/@stripo/core/files/710.4.5/build-project.js

Tchoupinax:
Could it be a good first issue to participate? If yes, I'm in!

fraxken:
@Tchoupinax The issue is not much complicated but I think we would run into some technical problems. I created this issue that we need to solve before this one: https://github.com/NodeSecure/js-x-ray/issues/353

That's an ok task for a first contribution

fraxken:
https://github.com/DataDog/guarddog/blob/main/guarddog/analyzer/sourcecode/npm-exfiltrate-sensitive-data.yml

fraxken:
https://socket.dev/npm/package/cms-ui-components/files/99.99.9/install.js

fraxken:
Closing the issue since the feature has been implemented (further enhancement will be spec in new issues)

PR #399, opened 2025-07-18
feat(probes): add minimal implementation of data exfiltration

clemgbld:
#279

### DIFF
--- a/workspaces/js-x-ray/src/probes/data-exfiltration.ts
+++ b/workspaces/js-x-ray/src/probes/data-exfiltration.ts
@@ -1,0 +1,111 @@
+// Import Third-party Dependencies
+import {
+  getCallExpressionIdentifier
+} from "@nodesecure/estree-ast-utils";
+import type { ESTree } from "meriyah";
+
+// Import Internal Dependencies
+import { generateWarning } from "../warnings.js";
+import type { ProbeContext } from "../ProbeRunner.js";
+import { rootLocation, toArrayLocation, type SourceArrayLocation } from "../utils/toArrayLocation.js";
+
+// CONSTANTS
+const kSensitiveMethods = [
+  "os.userInfo",
+  "os.networkInterfaces",
+  "os.cpus",
+  "dns.getServers"
+];
+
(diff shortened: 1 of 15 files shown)


workspaces/js-x-ray/src/probes/data-exfiltration.ts
```diff
@@ -0,0 +1,108 @@
...
+    moduleName: "os",
+    followConsecutiveAssignment: true
+  }).trace("os.cpus", {
+    moduleName: "os",
+    followConsecutiveAssignment: true
+  })
+    .trace("dns.getServers", {
+      moduleName: "dns",
+      followConsecutiveAssignment: true
+    });
```
fraxken: ? Syntax issue?

workspaces/js-x-ray/src/probes/data-exfiltration.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+function finalize(ctx: ProbeContext) {
+  const { sourceFile, context } = ctx;
+  if (context && Object.keys(context).length > 0) {
+    const warning = generateWarning("data-exfiltration",
+      { value: Object.keys(context).join(", ") });
+    sourceFile.warnings.push({ ...warning, location: Object.values(context).flat() });
+    Object.keys(context).forEach((key) => delete context[key]);
```
fraxken: Is this really required? The memory will be free by GC anyway so I don't think that useful

workspaces/js-x-ray/src/probes/data-exfiltration.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  return [true];
+}
+
+function main(
+  node: ESTree.Node,
+  ctx: ProbeContext
+) {
+  const castedNode = node as ESTree.CallExpression;
```
fraxken: Not sure why you need to cast?

```suggestion
function main(
  node: ESTree.CallExpression,
  ctx: ProbeContext
) {
  const castedNode = node;
```

fraxken (approved):

fraxken:
We also need to add the warning in README and docs

fraxken (approved):

PR #414, opened 2025-11-07
refactor: implement isStringBase64 and remove unmaintained dep

fraxken:


workspaces/sec-literal/test/isStringBase64.spec.ts
```diff
@@ -0,0 +1,50 @@
+/* eslint-disable @stylistic/max-len */
+// Import Node.js Dependencies
+import { test } from "node:test";
+import assert from "node:assert/strict";
+
+// Import Internal Dependencies
+import { isStringBase64 } from "../src/utils.js";
+
+test("isBase64", function isBase64() {
```
tony-go, line 9: I just would have split it in different test cases. Just in case of regression we could see which one fails

fraxken: That's the original test-suite of the package ^^, i'm quite ok with it since anyway stacktrace show which assertion fail.

tony-go: Yeah make sense

tony-go (approved):

PR #415, opened 2025-11-07
refactor(js-x-ray): remove is-minified-code from dependencies and re-implement it in utils

fraxken:


workspaces/js-x-ray/src/utils/isMinifiedCode.ts
```diff
@@ -0,0 +1,32 @@
+// CONSTANTS
+const kCommentPattern = /\/\*[\s\S]*?\*\/\r?\n?|\/\/.{0,200}?(?:\r?\n|$)/g;
+const kTrailingLfPattern = /\r?\n$/;
+
+/**
+ * This code has been imported from:
+ * https://github.com/MartinKolarik/is-minified-code
+ */
+export function isMinifiedCode(
```
PierreDemailly, line 9: Shouldn't we also import some unit tests, or write our own?

PierreDemailly (approved):

workspaces/js-x-ray/src/utils/isMinifiedCode.ts
fraxken: A bit complicated because how they are done in the package itself (he download 10k files from npm).

My overall idea was ok the current test suite already test it a bit and this function is not going to be updated till the day we replace it by maybe by something else.

PR #438, opened 2025-12-23
feat(EntryFileAnalyser): implement TypeScript files support

fraxken:


workspaces/js-x-ray/src/EntryFilesAnalyser.ts
```diff
@@ -28,6 +37,11 @@ export interface EntryFilesAnalyserOptions {
 }
 
 export class EntryFilesAnalyser {
+  static Parsers = {
+    js: new JsSourceParser(),
+    ts: new TsSourceParser() as unknown as SourceParser
```
clemgbld, line 42: I wonder if it’s possible to share a type file between the two workspaces where we would put the SourceParser interface and make the TsSourceParser implement it

PierreDemailly (approved):

Issue #460, opened 2026-01-15
(Probe) support multiple named main entrypoints with a required default


fraxken:
Right now, a probe can define multiple `validateNode` functions but only a single `main` function, and there is no way to correlate the two.

My idea is to introduce **named main handlers**:

```ts
export default {
  name: "isSerializeEnv",
  validateNode,
  initialize,
  main: {
    isEnv,
    isJSONStringifyEnv,
    default: isJSONStringifyEnv
  },
  breakOnMatch: false
};
```

> [!NOTE]
> Named `main` handlers must always provide a `default`.

Then, from within `validateNode` (via `this` or through `ProbeContext`), we can select which named `main` handler should be invoked. If nothing is configured—or if the provided name does not exist—the `default` handler is used.

```js
this.setEntryPoint("abc");
```

If a probe does not export named `main` handlers, calling `setEntryPoint()` has no effect.

7amed3li:
@fraxken  Got it, thanks for the detailed proposal.

I’ll take some time today/tomorrow to go through the code and sketch a clean implementation plan. I’ll open a PR once I’m confident with the approach.

PR #462, opened 2026-01-16
feat: support multiple named main handlers in probes (#460)

7amed3li:
## Description

This PR introduces the core infrastructure for supporting multiple named `main` handlers in probes, addressing issue #460.

It enables probes to correlate specific validation scenarios with distinct execution paths, while preserving full backward compatibility with the existing single-function `main` API.

## Architectural Changes

- Introduced the `NamedMainHandlers` type to allow probes to expose multiple execution entry points with a mandatory `default` handler.
- Extended `ProbeContext` with a `setEntryPoint(handlerName)` API, enabling handler selection during the validation phase.
- Updated the `Probe` interface to support either a legacy single `main` function or a `NamedMainHandlers` object.

## Implementation Details

- **Handler Resolution:** Updated `ProbeRunner#runProbe` to resolve and invoke the appropriate handler based on the selected entry point, falling back to `default` when needed.
- **State Management:** Added internal tracking for selected entry points per probe, ensuring the state is cleared after each invocation to prevent cross-node leakage.
- **Validation:** Enforced constructor-level validation to guarantee that probes using named handlers always provide a `default` implementation.

## Tests & Maintenance

- Refactored `test/ProbeRunner.spec.ts` by introducing shared assertion helpers (`assertProbeCtx`, `assertProbeMainContext`) to reduce duplication and improve readability.
- Removed scattered TypeScript suppressions by consolidating necessary casting inside test helpers.
- Resolved existing TypeScript warnings related to tuple access in mocked calls.

## Verification

- **Backward Compatibility:** All existing tests pass without modification.
- **Coverage:** Added targeted tests covering entry point selection, default fallback behavior, and edge cases.
- **Type Safety:** Verified a clean TypeScript build using `tsc --noEmit`.

### DIFF
--- a/workspaces/js-x-ray/src/ProbeRunner.ts
+++ b/workspaces/js-x-ray/src/ProbeRunner.ts
@@ -155,7 +155,6 @@ export class ProbeRunner {
       if (typeof probe.main === "object") {
         this.#selectedEntryPoints.set(probe, handlerName);
       }
-      // If probe.main is a function, this call has no effect
     };
 
     return {


7amed3li:
This PR focuses on the core infrastructure only.

If this design looks good, I can follow up with a separate PR demonstrating usage on an existing probe.

fraxken:
I'm off for few days, i will review tuesday or friday

workspaces/js-x-ray/src/ProbeRunner.ts
```diff
@@ -162,7 +188,27 @@ export class ProbeRunner {
       );
 
       if (isMatching) {
-        return probe.main(node, {
+        // Resolve which main handler to invoke
+        let mainHandler: (node: any, ctx: ProbeMainContext) => ProbeReturn;
+
+        if (typeof probe.main === "function") {
+          // Legacy single function
```
clemgbld: The comment is kind of misleading since the single function is not legacy it just a way to pass a single function when you don't need multiple
```suggestion
```

workspaces/js-x-ray/src/ProbeRunner.ts
```diff
@@ -140,10 +157,19 @@ export class ProbeRunner {
   #getProbeContext(
     probe: Probe
   ): ProbeContext {
+    const setEntryPoint = (handlerName: string) => {
+      // Only set entry point if probe has named handlers
```
clemgbld: I would remove the comments, i find the code explicit enough already

clemgbld:
We can take the opportunity to update the documentation about custom probe as well /workspaces/js-x-ray/docs/AstAnalyser.md

workspaces/js-x-ray/test/ProbeRunner.spec.ts
```diff
@@ -13,6 +13,31 @@ import {
...
+  context?: any;
+  data?: any;
+  signals: any;
+}) {
+  const c = ctx as any;
+  assertProbeCtx(c, expected);
+  assert.deepStrictEqual(c.data, expected.data);
+  assert.strictEqual(c.signals, expected.signals);
+}
+
```
clemgbld, line 40: I guess you could make a test to verify that you can indeed switch of entry point by doing a fake probe that call setEntryPoint and verify that the entry point is indeed cleared

7amed3li: > I guess you could make a test to verify that you can indeed switch of entry point by doing a fake probe that call setEntryPoint and verify that the entry point is indeed cleared

Good point 👍  
I’ve added a unit test covering `setEntryPoint`, verifying that it correctly switches the handler for one execution and is cleared afterward.  
No source changes, tests are passing.

workspaces/js-x-ray/src/ProbeRunner.ts
```diff
@@ -49,10 +57,8 @@ export interface Probe<T extends ProbeContextDef = ProbeContextDef> {
   initialize?: (ctx: ProbeContext<T>) => void | ProbeContext;
   finalize?: (ctx: ProbeContext<T>) => void;
   validateNode: ProbeValidationCallback<T> | ProbeValidationCallback<T>[];
-  main: (
-    node: any,
-    ctx: ProbeMainContext<T>
-  ) => ProbeReturn;
+  // Support both single function and named handlers
```
fraxken: Remove this comment (provide no value)

workspaces/js-x-ray/src/ProbeRunner.ts
```diff
@@ -30,10 +30,18 @@ const kProbeOriginalContext = Symbol.for("ProbeOriginalContext");
 
 export type ProbeReturn = void | null | symbol;
 export type ProbeContextDef = Record<string, any>;
+
+// Named main handlers type for probes with multiple entry points
```
fraxken: Remove this comment (provide no value)

workspaces/js-x-ray/src/ProbeRunner.ts
```diff
@@ -63,6 +69,9 @@ export class ProbeRunner {
   probes: Probe[];
   sourceFile: SourceFile;
   #collectableSetRegistry: CollectableSetRegistry;
+  // Store selected entry point per probe
+  // CRITICAL: Cleared after each invocation to avoid bleed between nodes
+  #selectedEntryPoints: Map<Probe, string> = new Map();
```
fraxken: Remove this comment (provide no value)

workspaces/js-x-ray/src/ProbeRunner.ts
```diff
@@ -108,10 +117,18 @@ export class ProbeRunner {
         typeof probe.validateNode === "function" || Array.isArray(probe.validateNode),
         `Invalid probe ${probe.name}: validateNode must be a function or an array of functions`
       );
+      // Validate main: either function or object with named handlers
```
fraxken: Remove this comment (provide no value)

workspaces/js-x-ray/src/ProbeRunner.ts
```diff
@@ -108,10 +117,18 @@ export class ProbeRunner {
...
       assert(
-        typeof probe.main === "function",
-        `Invalid probe ${probe.name}: main must be a function`
+        typeof probe.main === "function" || typeof probe.main === "object",
+        `Invalid probe ${probe.name}: main must be a function or an object with named handlers`
       );
+      // If named handlers, ensure 'default' handler exists
```
fraxken: Remove this comment (provide no value)

workspaces/js-x-ray/src/ProbeRunner.ts
```diff
@@ -162,7 +187,26 @@ export class ProbeRunner {
...
+            : "default";
+          mainHandler = probe.main[handlerName];
+        }
+
+        // CRITICAL: Clear entry point immediately after resolution
+        // This prevents bleed between different nodes in the same file
+        this.#selectedEntryPoints.delete(probe);
+
+        // Invoke handler with same context/data as legacy main
+        return mainHandler(node, {
```
fraxken: Remove all comments here too

fraxken: You can eventually keep the one for the selectedEntryPoints.delete

workspaces/js-x-ray/test/ProbeRunner.spec.ts
```diff
@@ -221,6 +265,39 @@ describe("ProbeRunner", () => {
       assert.strictEqual(result, "skip");
       assert.strictEqual(fakeProbe.teardown.mock.calls.length, 1);
     });
+    it("should trigger setEntryPoint and be cleared after execution", () => {
```
fraxken, line 269: Add new line before this test

7amed3li:
@fraxken  Thanks for getting back to this and for the review ^-^

Quick question for my own learning: do you generally prefer comments only when they explain non-obvious behavior or important side effects, and avoid comments that restate what the code already shows?

workspaces/js-x-ray/src/ProbeRunner.ts
```diff
@@ -140,10 +151,18 @@ export class ProbeRunner {
   #getProbeContext(
     probe: Probe
   ): ProbeContext {
+    const setEntryPoint = (handlerName: string) => {
+      if (typeof probe.main === "object") {
+        this.#selectedEntryPoints.set(probe, handlerName);
+      }
+      // If probe.main is a function, this call has no effect
```
clemgbld:
```suggestion
```
You can remove this comment as well, it's obvious if we look at line 155

fraxken:
@7amed3li Yes most of the time we expect comments for things we couldn't extract from the code by itself. LLMs tend to add extra noise that add no extra value by just reciting the code itself.

fraxken (approved):

Issue #473, opened 2026-01-22
Additional custom probes ideas


fraxken:
This issue is not a task by itself, the goal is to enumerate and discuss of good ideas of custom probes that the project could provide for a code audit.

# Crytographic usage
Flag any usage of Node.js crypto hash APIs.

- crypto.argon2
- crypto.bcrypt
- crypto.scrypt

# Path traversal
TBD (Not sure how)

PR #479, opened 2026-01-24
feat(js-x-ray): implement log-usage probe

clemgbld:
#473


workspaces/js-x-ray/src/probes/log-usage.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+function initialize(
+  ctx: ProbeContext
+) {
+  const { sourceFile } = ctx;
+
+  sourceFile.tracer.trace("console.log", {
+    followConsecutiveAssignment: true
+  });
```
fraxken: Probably need all others method (such as error, debug, trace)

PR #485, opened 2026-01-26
feat: implement new default isMonkeyPatch probe with monkey-patch warning

fraxken:


clemgbld:
Regarding the name of the probe "isMonkeyPatch" since data-exfiltration i started to use kebab-case for the naming of the file and of the probe on your recommandation, we should probably settle on one or the other naming in the futur for consistency

PR #496, opened 2026-01-28
feat: implement benchmarks using mitata

### LINKED ISSUE #490: Implement benchmarks with mitata
We need to implement benchmarks with `mitata` (probably using few representative codes) to have a performance baseline to work on performance improvment.

Right now we have nothing so it's hard to tell if a given update improve or degrade performance.

7amed3li:
**Adds mitata benchmarks using representative examples to establish a performance baseline.**

Closes #490

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,7 @@ $ yarn add @nodesecure/estree-ast-util
 
 ## Benchmarks
 
-This utility helps measure and track the performance of `js-x-ray` using [mitata](https://github.com/evanwashere/mitata).
+The performance of js-x-ray is measured and tracked using [mitata](https://github.com/evanwashere/mitata).
 
 To run the benchmarks:
 1. Navigate to `workspaces/js-x-ray`.


fraxken:
- No need to add a new tsconfig.json
- tsx is not required (we are using Node.js type stripping).
- Please make sure ESLint is happy

workspaces/js-x-ray/benchmark/fixtures.ts
```diff
@@ -0,0 +1,35 @@
...
+  ? import.meta.dirname
+  : fileURLToPath(new URL(".", import.meta.url));
+
+const EXAMPLES_DIR = resolve(__dirname, "../examples");
+
+export function loadFixtures(): Fixture[] {
+  const files = readdirSync(EXAMPLES_DIR);
+
+  return files
+    .filter((file) => file.endsWith(".js"))
```
clemgbld: I think we can remove that line since every example are in js (and will be, if not it will be ts or jsx which is fine)

workspaces/js-x-ray/benchmark/fixtures.ts
```diff
@@ -0,0 +1,35 @@
...
+      const path = join(EXAMPLES_DIR, file);
+      const content = readFileSync(path, "utf-8");
+
+      return {
+        name: file,
+        content,
+        size: content.length
+      };
+    })
+    .sort((a, b) => a.size - b.size);
```
clemgbld: Not really sure why are we sorting the files here ?

workspaces/js-x-ray/benchmark/index.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  if (mediumFixture) {
+    bench(`Medium File (${mediumFixture.name} - ${(mediumFixture.size / 1024).toFixed(2)}KB)`, () => {
+      analyser.analyse(mediumFixture.content);
+    });
+  }
+
+  if (largeFixture) {
+    bench(`Large File (${largeFixture.name} - ${(largeFixture.size / 1024).toFixed(2)}KB)`, () => {
```
clemgbld: a function to factor out that formatting logic would be nice:
```js
`Large File ${formatFileSize(largeFixture)})`
```

workspaces/js-x-ray/benchmark/index.ts
```diff
@@ -0,0 +1,50 @@
...
+
+// Import Internal Dependencies
+import { AstAnalyser } from "../src/index.ts";
+import { loadFixtures } from "./fixtures.ts";
+
+const fixtures = loadFixtures();
+const analyser = new AstAnalyser();
+
+group("AstAnalyser.analyse()", () => {
+  const smallFixture = fixtures.find((f) => f.name === "jquery.min.js") || fixtures[0];
```
clemgbld: couldn't we target smallFixture, mediumFixture and largeFixture by size instead of hardcoding the names ?
it would be tedious to have to modify the benchmark each time that we add an example

workspaces/js-x-ray/benchmark/fixtures.ts
```diff
@@ -0,0 +1,35 @@
+
+// Import Node.js Dependencies
+import { readFileSync, readdirSync } from "node:fs";
```
clemgbld: In my opinion we could just put that function into the index.ts file directly

workspaces/js-x-ray/benchmark/README.md
```diff
@@ -0,0 +1,30 @@
+# Benchmarks
+
+This utility helps measure and track the performance of `js-x-ray` using [mitata](https://github.com/evanwashere/mitata).
```
clemgbld: I think we should mention the benchmark in the README and not here

workspaces/js-x-ray/benchmark/index.ts
```diff
@@ -0,0 +1,86 @@
...
+// Import Third-party Dependencies
+import { run, bench, group } from "mitata";
+
+// Import Internal Dependencies
+import { AstAnalyser } from "../src/index.ts";
+
+// Constants
+const __dirname = typeof import.meta.dirname === "string"
+  ? import.meta.dirname
+  : fileURLToPath(new URL(".", import.meta.url));
```
fraxken: Just use import.meta.dirname (there is no need for a fallback)

README.md
```diff
@@ -115,6 +115,14 @@ Alternatively, you can use `EntryFilesAnalyser` directly for multi-file analysis
 - [EntryFilesAnalyser](./workspaces/js-x-ray/docs/EntryFilesAnalyser.md)
 - [CollectableSet](./workspaces/js-x-ray/docs/CollectableSet.md)
 
+## Benchmarks
+
+This utility helps measure and track the performance of `js-x-ray` using [mitata](https://github.com/evanwashere/mitata).
+
+To run the benchmarks:
+1. Navigate to `workspaces/js-x-ray`.
+2. Run `npm run bench`.
```
fraxken: I would prefer this section close to the end of README (after workspaces) since this is more oriented for contributors than developers using the package itself.

workspaces/js-x-ray/benchmark/index.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  for (const fixture of obfuscated) {
+    bench(`${fixture.name}`, () => {
+      analyser.analyse(fixture.content);
+    });
+  }
+});
+
+await run({
+  colors: true
```
clemgbld: color is already true by default since we didn't set up any env to disable it:
https://github.com/evanwashere/mitata/blob/master/src/main.mjs#L231

README.md
```diff
@@ -197,6 +198,14 @@ $ npm i @nodesecure/estree-ast-util
 $ yarn add @nodesecure/estree-ast-util
 ```
 
+## Benchmarks
+
+This utility helps measure and track the performance of `js-x-ray` using [mitata](https://github.com/evanwashere/mitata).
```
clemgbld:
```suggestion
The performance of js-x-ray is measured and tracked using [mitata](https://github.com/evanwashere/mitata).
```

workspaces/js-x-ray/benchmark/index.ts
```diff
@@ -0,0 +1,80 @@
...
+const fixtures = loadFixtures();
+const analyser = new AstAnalyser();
+
+function formatFileSize(fixture: Fixture): string {
+  return `${fixture.name} - ${(fixture.size / 1024).toFixed(2)}KB`;
+}
+
+group("AstAnalyser.analyse()", () => {
+  const sortedFixtures = [...fixtures].sort((a, b) => a.size - b.size);
+  const smallFixture = sortedFixtures[0];
```
clemgbld: We should have some constants who represent the threshold of what is a small file vs a medium file vs a large file, and filter according to that.

7amed3li:
Thanks for the feedback , 
 I applied the changes 👍

workspaces/js-x-ray/benchmark/index.ts
```diff
@@ -0,0 +1,84 @@
...
+const fixtures = loadFixtures();
+const analyser = new AstAnalyser();
+
+function formatFileSize(fixture: Fixture): string {
+  return `${fixture.name} - ${(fixture.size / 1024).toFixed(2)}KB`;
+}
+
+group("AstAnalyser.analyse()", () => {
+  const sortedFixtures = [...fixtures].sort((a, b) => a.size - b.size);
+  const smallFixture = sortedFixtures.find((f) => f.size < SMALL_THRESHOLD) || sortedFixtures[0];
```
clemgbld:
```suggestion
  const smallFixture = sortedFixtures.find((f) => f.size < SMALL_THRESHOLD);
```

clemgbld: should be an array of small files

workspaces/js-x-ray/benchmark/index.ts
```diff
@@ -0,0 +1,84 @@
...
+  const mediumFixture = sortedFixtures.find((f) => f.size >= SMALL_THRESHOLD && f.size < LARGE_THRESHOLD);
```
clemgbld: should be array of medium files

workspaces/js-x-ray/benchmark/index.ts
```diff
@@ -0,0 +1,84 @@
...
+  const largeFixture = sortedFixtures.find((f) => f.size >= LARGE_THRESHOLD) || sortedFixtures[sortedFixtures.length - 1];
```
clemgbld: should be an array of large files

PR #537, opened 2026-02-18
test(utils): add unit tests for stripNodePrefix

bashlor:
Adds unit tests for stripNodePrefix covering:
- valid "node:" prefix removal
- non-prefixed values
- similar but invalid prefixes
- non-string inputs

### DIFF
--- a/workspaces/js-x-ray/test/utils/stripNodePrefix.spec.ts
+++ b/workspaces/js-x-ray/test/utils/stripNodePrefix.spec.ts
@@ -4,6 +4,7 @@ import { describe, it } from "node:test";
 
 // Import Internal Dependencies
 import { stripNodePrefix } from "../../src/utils/index.ts";
+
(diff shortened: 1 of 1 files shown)


workspaces/js-x-ray/test/utils/stripNodePrefix.spec.ts
```diff
@@ -0,0 +1,38 @@
+// Import Node.js Dependencies
+import assert from "node:assert";
+import { describe, it } from "node:test";
+
+// Import Internal Dependencies
+import { stripNodePrefix } from "../../src/utils/index.ts";
+describe("stripNodePrefix", () => {
```
PierreDemailly, line 8:
```suggestion
import { stripNodePrefix } from "../../src/utils/index.ts";

describe("stripNodePrefix", () => {
```

Issue #506, opened 2026-01-30
Optional warning for crypto API misuse detection


fraxken:
Implement a new optional warning and probe to detect usage of Node.js crypto APIs that could lead to security vulnerabilities, particularly around incorrect hash generation and weak cryptographic practices.

**Proposed approach:**

Add detection for potentially insecure usage of the following crypto methods:

**Password hashing functions:**
- [x] `crypto.argon2()`
- [x] `crypto.bcrypt()`
- [x] `crypto.scrypt()`
- [ ] `crypto.pbkdf2()` / `crypto.pbkdf2Sync()`

**Insecure random generation:**
- [x] `Math.random()` used for security-sensitive operations (instead of `crypto.randomBytes()`)

_Note: This list should be expanded by reviewing the Node.js crypto documentation to ensure comprehensive coverage._

**Examples:**

```javascript
// Should trigger warning - insufficient iterations for pbkdf2
crypto.pbkdf2(password, salt, 1000, 64, 'sha512', callback);

// Should trigger warning - insecure random for tokens
const token = Math.random().toString(36);

// Should NOT trigger warning - proper usage
crypto.scrypt(password, salt, 64, (err, derivedKey) => {
  // proper implementation
});
```

**Detection criteria:**

The probe should identify:
1. Insufficient iteration counts for key derivation functions
2. Missing or weak salt generation
3. Use of non-cryptographic random functions for security contexts

**Expected behavior:**

Emit warnings with severity levels:
- **Warning** Weak parameters (low iteration counts)
- **Information** Potential misuse requiring manual review

PR #578, opened 2026-03-22
feat(js-x-ray): add weak-scrypt detection probe

HoyeongJeon:
Adds an optional weak-scrypt probe to detect insecure usage of crypto.scrypt().

 Detection targets:                                                                 
  - Hardcoded salt (string literal as salt argument) — salt must be randomly
  generated per NIST SP 800-132                                      
  - Short salt (string literal less than 16 characters) — minimum 128-bit salt per
  NIST SP 800-132
  - Insufficient cost parameter (below 16384) — Node.js default and OWASP minimum
  recommendation
                                                                                     
Partially addresses #506 (scrypt part)


workspaces/js-x-ray/src/probes/isWeakScrypt.ts
```diff
@@ -0,0 +1,94 @@
...
+  const { sourceFile } = ctx;
+  const salt = node.arguments.at(1);
+  const options = node.arguments.at(3);
+
+  if (options && options.type === "ObjectExpression") {
+    for (const p of options.properties) {
+      if (
+        p.type === "Property"
+        && p.key.type === "Identifier"
+        && p.key.name === "cost"
```
PierreDemailly: Should we also check the prop `N` which is an alias for `cost`?

```suggestion
        && (p.key.name === "cost" || p.key.name === "N")
```

PierreDemailly: Should we also check `N` which is an alias for `cost` ?

```suggestion
        && (p.key.name === "cost" || p.key.name === "N")
```

workspaces/js-x-ray/src/probes/isWeakScrypt.ts
```diff
@@ -0,0 +1,94 @@
...
+        && p.value.type === "Literal"
+        && typeof p.value.value === "number"
+        && p.value.value < 16384
```
PierreDemailly: I'm not confident with 16384 as a recommended minimum cost.
https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#scrypt
The minimal cost recommendation depends on parallelization/p.                                      
                                                                                                                     
>N=2^17 (128 MiB), r=8 (1024 bytes), p=1                                                                          
N=2^16 (64 MiB), r=8 (1024 bytes), p=2                                                                           
N=2^15 (32 MiB), r=8 (1024 bytes), p=3                                                                           
N=2^14 (16 MiB), r=8 (1024 bytes), p=5
N=2^13 (8 MiB), r=8 (1024 bytes), p=10                                                                           
                                                
Node.js default blocksize is r=8 & default parallelization is p=1.                                                   
With these values, its default cost is well below the OWASP minimum recommendation.

I think we should base the warning on all three parameters (cost, blockSize, parallelization) rather than cost alone

HoyeongJeon:
Thank you for the review, @PierreDemailly!                                                                                 I carefully reviewed the document you mentioned and rebuilt the validation logic to check all three parameters (N, r, p) against the OWASP recommended minimums, along with supporting their aliases

workspaces/js-x-ray/src/probes/isWeakScrypt.ts
```diff
@@ -0,0 +1,158 @@
...
+import { generateWarning } from "../warnings.ts";
+
+/**
+ * OWASP recommended minimum scrypt parameter combinations.
+ * Each entry is [minN, minP] — sorted by N descending.
+ * All recommendations assume r >= 8.
+ *
+ * @see https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#scrypt
+ */
+const OWASP_MIN_PARAMS: [number, number][] = [
```
PierreDemailly:
```suggestion
const OWASP_MIN_PARAMS: [cost: number, parallelization: number][] = [
```

workspaces/js-x-ray/src/probes/isWeakScrypt.ts
```diff
@@ -0,0 +1,158 @@
...
+  [16384, 5],
+  [8192, 10]
+];
+
+const MIN_BLOCK_SIZE = 8;
+
+// Node.js crypto.scrypt defaults
+const DEFAULT_COST = 16384;
+const DEFAULT_BLOCK_SIZE = 8;
+const DEFAULT_PARALLELIZATION = 1;
```
PierreDemailly:
```suggestion
const kOWASPMinParams: [number, number][] = [
  [131072, 1],
  [65536, 2],
  [32768, 3],
  [16384, 5],
  [8192, 10]
];

const kMinBlockSize = 8;

// Node.js crypto.scrypt defaults
const kDefaultCost = 16384;
const kDefaultBlockSize = 8;
const kDefaultParallelization = 1;
```

`CONSTANT_CASE` is reserved to exported constants, more info [here](https://github.com/OpenAlly/configs/blob/main/src/eslint/docs/constants.md)

PierreDemailly (approved):
lgtm

workspaces/js-x-ray/src/probes/isWeakScrypt.ts
HoyeongJeon: Thank you for the detail!

workspaces/js-x-ray/src/probes/isWeakScrypt.ts
HoyeongJeon: I missed the doc 😓
Thank you for the info

workspaces/js-x-ray/src/probes/isWeakScrypt.ts
```diff
@@ -0,0 +1,158 @@
...
+      typeof prop.value.value === "number"
+    ) {
+      return prop.value.value;
+    }
+  }
+
+  return null;
+}
+
+function isWeakScryptParams(N: number, r: number, p: number): boolean {
```
fraxken: can we avoid one letter arguments?

fraxken:
Think to update documentations

HoyeongJeon:
Rebased on upstream/master to sync with latest changes
@fraxken Should I update more documentations? I'm a bit confused about which documentations did you mean

fraxken:
@HoyeongJeon You also need to update the README with the new warning

HoyeongJeon:
@fraxken Let me know if there are any other changes needed

clemgbld (approved):

fraxken (approved):

fraxken:
@allcontributors please add @HoyeongJeon for code

PR #592, opened 2026-03-28
feat(js-x-ray): add weak-argon2 detection probe

HoyeongJeon:
Adds an optional weak-argon2 probe to detect insecure usage of crypto.argon2() / crypto.argon2Sync().                                                                                                    

Detection targets:
  - Wrong algorithm (argon2d or argon2i instead of argon2id) — OWASP recommends argon2id as it    combines side-channel resistance and GPU cracking resistance
  - Weak parameters (memory/passes/parallelism below OWASP minimum) — validates against five OWASP-recommended parameter combinations trading memory for iterations
  - Hardcoded nonce (string literal as nonce argument) — salt must be randomly generated, not hardcoded                                                                                         

Partially addresses #506 (Argon2 part)


workspaces/js-x-ray/src/probes/isWeakArgon2.ts
```diff
@@ -0,0 +1,129 @@
...
+}
+
+function main(node: ESTree.CallExpression, ctx: ProbeContext) {
+  const { sourceFile } = ctx;
+  const algorithm = node.arguments.at(0);
+  if (isLiteral(algorithm)) {
+    if (algorithm.value !== "argon2id") {
+      sourceFile.warnings.push(
+        generateWarning("weak-argon2", {
+          value: "wrong-algorithm",
```
clemgbld: i guess it would make the warning more useful if the value is the wrong algorithm itself ```algorithm.value ```

clemgbld: or some think like ``` wrong-algorithm : algorithm.value ```

workspaces/js-x-ray/src/probes/isWeakArgon2.ts
```diff
@@ -0,0 +1,129 @@
...
+      followConsecutiveAssignment: true,
+      moduleName: "crypto"
+    });
+  }
+}
+
+function main(node: ESTree.CallExpression, ctx: ProbeContext) {
+  const { sourceFile } = ctx;
+  const algorithm = node.arguments.at(0);
+  if (isLiteral(algorithm)) {
```
clemgbld, line 89: the isLiteral case will be the most common one for sure, but imo we should also check the case where it is an identifier:

```js
const algo = "argon2d";

crypto.argon2(algo, {
  message: "password",
  nonce: crypto.randomBytes(16),
  memory: 47104,
  passes: 1,
  parallelism: 1,
  tagLength: 64
});

```

docs/weak-argon2.md
```diff
@@ -0,0 +1,49 @@
+# Weak argon2
+
+| Code | Severity | i18n | Experimental |
+| --- | --- | --- | :-: |
+| weak-argon2 | `Warning` | `sast_warnings.weak_argon2` | :x: |
+
+## Introduction
+
+Detect usage of **weak Argon2** parameters with the Node.js core `crypto.argon2()` / `crypto.argon2Sync()` functions. This probe checks for:
```
clemgbld, line 9:
```suggestion
Detect usage of **weak Argon2** parameters with the Node.js core `crypto.argon2()` / `crypto.argon2Sync()` functions. 
Checks for:
```

I don't think we should use the term "probe" since it is an implementation detail of Js-x-ray

workspaces/js-x-ray/src/probes/isWeakArgon2.ts
```diff
@@ -0,0 +1,129 @@
...
+          })
+        );
+      }
+    }
+
+    const nonce = properties.find(
+      (prop) => prop.key.type === "Identifier" && prop.key.name === "nonce"
+    );
+
+    if (nonce && isLiteral(nonce.value)) {
```
clemgbld, line 124: same here we should check for identifier that are hardcoded

workspaces/js-x-ray/src/probes/isWeakArgon2.ts
```diff
@@ -0,0 +1,129 @@
...
+      );
+    }
+  }
+
+  const parameters = node.arguments.at(1);
+  if (parameters && parameters.type === "ObjectExpression") {
+    const properties = parameters.properties.filter(
+      (prop): prop is ESTree.Property => prop.type === "Property"
+    );
+    const memory = extractNumericParam(properties, ["memory"]);
```
clemgbld, line 105: i guess a better way to do it would be to extract each properties in one loop then extract the numeric value

workspaces/js-x-ray/src/probes/isWeakArgon2.ts
```diff
@@ -0,0 +1,129 @@
...
+
+const tracedFunctions = new Set(["crypto.argon2", "crypto.argon2Sync"]);
+
+function extractNumericParam(
+  properties: ESTree.Property[],
+  names: string[]
+): number | null {
+  for (const prop of properties) {
+    if (
+      prop.key.type === "Identifier" &&
```
clemgbld, line 26: same here, it would be nice to check for identifier that are literal as well

workspaces/js-x-ray/src/probes/isWeakArgon2.ts
clemgbld: you can achieve that like this:

```js
// when the type is "Identifier"
const literalIdentifier = tracer.literalIdentifiers.get(node.name);
```

workspaces/js-x-ray/src/probes/isWeakArgon2.ts
```diff
@@ -0,0 +1,129 @@
...
+
+const kOWASPMinParams: [minMemory: number, minIteration: number, minParallelism: number][] = [
+  [47104, 1, 1],
+  [19456, 2, 1],
+  [12288, 3, 1],
+  [9216, 4, 1],
+  [7168, 5, 1]
+];
+
+const tracedFunctions = new Set(["crypto.argon2", "crypto.argon2Sync"]);
```
clemgbld, line 18: i guess, just creating an array is ok here, you already know that there will be no duplicate

workspaces/js-x-ray/src/probes/isWeakArgon2.ts
HoyeongJeon: > or some think like `wrong-algorithm : algorithm.value`

Thanks! I think it's more informative!

workspaces/js-x-ray/src/probes/isWeakArgon2.ts
```diff
@@ -0,0 +1,129 @@
...
+        sourceFile.warnings.push(
+          generateWarning("weak-argon2", {
+            value: "weak-parameters",
+            location: node.loc
+          })
+        );
+      }
+    }
+
+    const nonce = properties.find(
```
clemgbld, line 120: same here it should be part of just one loop rather than iterating once again

workspaces/js-x-ray/src/probes/isWeakArgon2.ts
```diff
@@ -0,0 +1,129 @@
...
+    );
+    const memory = extractNumericParam(properties, ["memory"]);
+    const iteration = extractNumericParam(properties, ["passes"]);
+    const parallelism = extractNumericParam(properties, ["parallelism"]);
+
+    if (memory && iteration && parallelism) {
+      if (isWeakArgon2Params(memory, iteration, parallelism)) {
+        sourceFile.warnings.push(
+          generateWarning("weak-argon2", {
+            value: "weak-parameters",
```
clemgbld, line 113: in the value it would be also more helpful to add the type of the parameters that is weak:
```value: `weak-parameter: `${type}`;

PR #626, opened 2026-04-24
feat(js-x-ray): improve unsafe-vm-context

### LINKED ISSUE #625: (Tracer) support tracing class instances's methods
Currently we’ve got not built in mechanism to easily trace class instance's methods.
This new tracer enhancement should be implemented without degrading the performance of Js-x-ray too much, running and inspecting the benchmark will help us make sure we respected this criteria.

This should be a pre-requisite to trace `runInContext`from the `vm.Script`: #599

clemgbld:
Close #625 
Close #624

### DIFF
--- a/workspaces/js-x-ray/src/probes/unsafe-vm-context.ts
+++ b/workspaces/js-x-ray/src/probes/unsafe-vm-context.ts
@@ -5,6 +5,10 @@ import type { ESTree } from "meriyah";
 import type { ProbeMainContext, ProbeContext } from "../ProbeRunner.ts";
 import { generateWarning } from "../warnings.ts";
 import { CALL_EXPRESSION_DATA } from "../contants.ts";
+import { VariableTracer, type ReturnValueEventPayload } from "../VariableTracer.ts";
+
+// CONSTANTS
+const kRunInContextTracedFunctions = Symbol("runInContextTracedFunctions");
 
 function validateNode(
   node: ESTree.Node,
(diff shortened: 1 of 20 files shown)


PierreDemailly (approved):

workspaces/js-x-ray/src/probes/unsafe-vm-context.ts
```diff
@@ -20,19 +24,47 @@ function validateNode(
...
     moduleName: "vm"
   });
+
+  tracer.trace("vm.Script", {
+    followReturnValueAssignement: true,
+    followConsecutiveAssignment: true,
+    moduleName: "vm"
+  });
+
+  tracer.on(VariableTracer.ReturnValueEvent, (payload: ReturnValueEventPayload) => {
```
fraxken, line 56: I wonder if there is a world where we could leak the listener between multiple run? I'm not 100% sure but I wonder if we should not make sure to clear that listener in finalize.

clemgbld: Normally a leak can't happen since listeners are cleared when the VariableTracer is garbage collected (and we instantiate a new VariableTracer in every run and never hold on to it).
But it could be nice to add this safeguard in case we change the implementation in the future.

clemgbld: i've done it directly in the ProbeRunner like that it covers every probes

fraxken (approved):

Issue #472, opened 2026-01-22
Log detection


fraxken:
Non-redacted log can often leak sensible informations (such as env, token, customer informations etc). The goal is to design a probe that could detect Console API usage (that we can expand to third-party logger in following PRs).

When a hit is detected we would throw an Information warning (such as `log-usage` or something similar).

This new warning should be optional (not included by default).

fraxken:
Next step would be to be able to identify logs from third-party logger such as pino

PR #638, opened 2026-05-14
feat(log-usage): add support for customLevels for pino

clemgbld:
#472

### DIFF
--- a/workspaces/js-x-ray/src/probes/log-usage.ts
+++ b/workspaces/js-x-ray/src/probes/log-usage.ts
@@ -9,7 +9,7 @@ import { toArrayLocation, type SourceArrayLocation } from "../utils/toArrayLocat
 import { VariableTracer, type ReturnValueEventPayload } from "../VariableTracer.ts";
 
 // CONSTANTS
-const kRunInContextTracedFunctions = Symbol("runInContextTracedFunctions");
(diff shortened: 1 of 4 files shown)


workspaces/js-x-ray/src/probes/log-usage.ts
```diff
@@ -46,12 +46,57 @@ function initialize(
...
     if (!pinoLoggerFactoryTracedFunctions.has(payload.name)) {
       return;
     }
 
-    for (const method of kPinoLogMethods) {
+    let pinoLoggerMethods: string[] = childLoggerFunctions.get(payload.name) ?? [...kPinoLogMethods];
+
+    if (payload.name === "pino") {
+      const pinoContext = payload.arguments[0];
+      if (pinoContext && pinoContext.type === "ObjectExpression") {
```
PierreDemailly:
```suggestion
    pino: if (payload.name === "pino") {
      const pinoContext = payload.arguments[0];
      if ((pinoContext?.type === "ObjectExpression") === false) {
        break pino;
      }
      // ...
```
nit: we could `break` the first condition to save1 indent level, wdyt?

PierreDemailly (approved):

fraxken (approved):

PR #661, opened 2026-06-29
Feat/unsafe prehash

ErwanRaulo:
Adds an optional unsafe-prehash probe to detect insecure usage of null byte enconding prehash, commonly used later on with bcrypt

Partially complete https://github.com/NodeSecure/js-x-ray/issues/506 for bcrypt (prehash is a subset)

### DIFF
--- a/workspaces/js-x-ray/src/probes/isUnsafePrehash.ts
+++ b/workspaces/js-x-ray/src/probes/isUnsafePrehash.ts
@@ -1,0 +1,250 @@
+// Import Third-party Dependencies
+import type { ESTree } from "meriyah";
+
+// Import Internal Dependencies
+import type { ProbeContext, ProbeMainContext } from "../ProbeRunner.ts";
+import { CALL_EXPRESSION_DATA } from "../contants.ts";
+import { isCallExpression, isLiteral } from "../estree/types.ts";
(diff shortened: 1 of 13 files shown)


workspaces/js-x-ray/src/probes/isUnsafePrehash.ts
```diff
@@ -0,0 +1,230 @@
...
+  hashNode: ESTree.Node | null | undefined
+): boolean {
+  const encodingArgs = resolveDigestEncodingArguments(hashNode);
+  if (encodingArgs === null) {
+    return false;
+  }
+
+  const encodingArg = encodingArgs.at(0);
+
+  return !(isLiteral(encodingArg) && kSafeDigestEncodings.has(encodingArg.value));
```
clemgbld: I think we should handle the case when the argument is an identifier that point to a literal

ErwanRaulo: done !

clemgbld (approved):

PierreDemailly (approved):

fraxken (approved):

fraxken:
@allcontributors please add @ErwanRaulo for code

PR #681, opened 2026-07-21
feat(i18n): add missing translations & export them for scanner reuse

ErwanRaulo:
few warnings lack of translations and in a global point of view all translations shoud be reused by tooling depending on js-x-ray vuln detection (like scanner)


clemgbld (approved):

workspaces/js-x-ray/package.json
```diff
@@ -12,6 +12,22 @@
...
+      "types": "./dist/i18n/french.d.ts"
+    },
+    "./i18n/arabic": {
+      "import": "./dist/i18n/arabic.js",
+      "types": "./dist/i18n/arabic.d.ts"
+    },
+    "./i18n/turkish": {
+      "import": "./dist/i18n/turkish.js",
+      "types": "./dist/i18n/turkish.d.ts"
+    },
```
fraxken: I think it support `*` pattern; See: https://github.com/NodeSecure/cli/blob/master/workspaces/documentation-ui/package.json#L30

fraxken: `"./i18n/*.js": "./dist/i18n/*.js"` should work here

ErwanRaulo: could not succeed using npm link. am i doing something wrong ?

ErwanRaulo: this format works well:

    "./i18n/*": {
      "types": "./dist/i18n/*.d.ts",
      "import": "./dist/i18n/*.js"
    },

fraxken (approved):

PR #699, opened 2026-08-30
feat(js-x-ray): add weak-argon2 detection probe

HoyeongJeon:
Adds an optional weak-argon2 probe to detect insecure usage of crypto.argon2() and crypto.argon2Sync().

Detection targets:
- argon2d variant — RFC 9106 §1 describes Argon2d as suitable for cryptocurrencies and
  proof-of-work applications with no threats from side-channel timing attacks; §4 recommends
  Argon2id when in doubt
- Insufficient memory/passes — parameters matching none of the five OWASP recommended combinations
- argon2i with fewer than 3 passes — OWASP marks the two lowest-pass rows "(Do not use with
  Argon2i)"; RFC 9106 §7.3 notes that 3 passes is almost optimal for Argon2i
- Hardcoded nonce (string literal) — RFC 9106 §3.1: the salt SHOULD be unique for each password
- Short nonce (string literal shorter than 16 characters) — RFC 9106 §3.1: 16 bytes is
  RECOMMENDED for password hashing

Identifiers assigned a literal are resolved through the `VariableTracer`, so
`const algo = "argon2d"; crypto.argon2(algo, ...)` is reported as well.

Parameters that cannot be resolved statically (variables, spread elements, keys computed from
variables) are skipped rather than reported. `parallelism` is not checked, since raising it does
not reduce the total memory cost.

Partially addresses #506 (argon2 part)

### DIFF
--- a/docs/crypto.weak-argon2.md
+++ b/docs/crypto.weak-argon2.md
@@ -1,0 +1,105 @@
+# Weak Argon2
+
+| Code | Severity | i18n | Experimental |
+| --- | --- | --- | :-: |
+| crypto.weak-argon2 | `Warning` | `sast_warnings.weak_argon2` | :white_check_mark: |
+
+## Introduction
+
+Detect usage of **weak Argon2** parameters with the Node.js core `crypto.argon2()` and `crypto.argon2Sync()` functions. This probe checks for:
+
+- **weak-algorithm**: the `argon2d` variant, which is data-dependent and not intended for password hashing.
+- **low-params**: `memory` and `passes` that do not meet [OWASP recommended combinations](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#argon2id).
+- **short-nonce**: nonce is a hardcoded string literal shorter than 16 characters.
+- **hardcoded-nonce**: nonce is a hardcoded string literal (should be randomly generated).
+
+> [!NOTE]
+> Argon2 calls the salt a **nonce**, and so does the Node.js API. The warning values follow that
+> naming so they match what the analysed source code actually writes.
+
+## Variants
+
+[RFC 9106](https://www.rfc-editor.org/rfc/rfc9106.html#section-4) recommends `argon2id` when in doubt, and describes `argon2d` as suited to applications with no threat from side-channel attacks — which excludes password hashing.
+
+| Variant | Verdict |
+| --- | --- |
+| `argon2d` | Always reported. Data-dependent memory access leaks information through side channels. |
+| `argon2i` | Accepted, but restricted to `passes >= 3` (see below). |
+| `argon2id` | Accepted. |
+
+## Parameters
+
+OWASP publishes five configurations that provide an equal level of defense, trading CPU against RAM. A call is reported when its `memory` and `passes` satisfy none of them.
+
+| memory (KiB) | passes | |
+| --- | --- | --- |
+| 47104 (46 MiB) | 1 | Do not use with Argon2i |
+| 19456 (19 MiB) | 2 | Do not use with Argon2i |
+| 12288 (12 MiB) | 3 | |
+| 9216 (9 MiB) | 4 | |
+| 7168 (7 MiB) | 5 | |
+
+Argon2i is data-independent and therefore weaker against time-memory trade-off attacks, which additional passes mitigate — see [RFC 9106 section 7.3](https://www.rfc-editor.org/rfc/rfc9106.html#section-7.3). The two lowest-pass rows are consequently unavailable to it, so `argon2i` with `passes` below 3 is reported regardless of how much memory is allocated.
+
+`parallelism` is not checked: raising it does not reduce the total memory cost, so it is not a weakness on its own.
+
+Parameters passed as variables are not resolved, and no warning is emitted for them.
+
+## Example
+
+```js
+import crypto from "crypto";
+
+// weak-algorithm: argon2d is not intended for password hashing
+crypto.argon2("argon2d", {
+  message: password,
+  nonce: crypto.randomBytes(16),
+  parallelism: 1,
+  tagLength: 32,
+  memory: 19456,
+  passes: 2
+}, (err, tag) => {});
+
+// low-params: 8 KiB is far below the lowest OWASP row (7168 KiB)
+crypto.argon2("argon2id", {
+  message: password,
+  nonce: crypto.randomBytes(16),
+  parallelism: 1,
+  tagLength: 32,
+  memory: 8,
+  passes: 1
+}, (err, tag) => {});
+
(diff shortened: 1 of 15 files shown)


workspaces/js-x-ray/src/probes/crypto/isWeakArgon2.ts
```diff
@@ -0,0 +1,238 @@
...
+        generateWarning("crypto.weak-argon2", {
+          value: `low-params: ${weakParam}`,
+          location: node.loc
+        })
+      );
+    }
+  }
+
+  if (nonce !== null) {
+    sourceFile.warnings.push(
```
clemgbld, line 122: In the case when we generate multiple warning for the same function call it would be better to generate only one aggregated warning a bit like we do in the finalize of the probe `data-exfiltration`

docs/crypto.weak-argon2.md
```diff
@@ -0,0 +1,117 @@
...
+
+| memory (KiB) | passes | |
+| --- | --- | --- |
+| 47104 (46 MiB) | 1 | Do not use with Argon2i |
+| 19456 (19 MiB) | 2 | Do not use with Argon2i |
+| 12288 (12 MiB) | 3 | |
+| 9216 (9 MiB) | 4 | |
+| 7168 (7 MiB) | 5 | |
+
+Argon2i is data-independent and therefore weaker against time-memory trade-off attacks, which additional passes mitigate — see [RFC 9106 section 7.3](https://www.rfc-editor.org/rfc/rfc9106.html#section-7.3). The two lowest-pass rows are consequently unavailable to it, so `argon2i` with `passes` below 3 is reported regardless of how much memory is allocated.
```
ErwanRaulo, line 43: should we really accepts argon2i when passes >= 3 ? @clemgbld

workspaces/js-x-ray/src/probes/crypto/isWeakArgon2.ts
```diff
@@ -0,0 +1,238 @@
...
+
+  return isStringLiteral(prop.key) ? prop.key.value : null;
+}
+
+/**
+ * Read a string out of a node, following identifiers assigned a string literal.
+ * TemplateLiteral assignments are ignored because the tracer stores them with
+ * their interpolations replaced by placeholders.
+ */
+function resolveString(
```
ErwanRaulo: please rebase on master, this function is now a shared helper.

workspaces/js-x-ray/src/probes/crypto/isWeakArgon2.ts
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  return null;
+}
+
+/**
+ * Read a number out of a node, following identifiers assigned a numeric
+ * literal. The tracer stores every literal as a string, hence the conversion.
+ */
+function resolveNumber(
```
ErwanRaulo: same, this function is now a shared helper on master

workspaces/js-x-ray/src/probes/crypto/isWeakArgon2.ts
ErwanRaulo: but add this limitation fix :

`return literal?.type === "Literal" ? literal.value : null;`

a test is also missing for this template literal guard.

ErwanRaulo: a test is also missing for this template literal guard.

workspaces/js-x-ray/src/probes/crypto/isWeakArgon2.ts
```diff
@@ -0,0 +1,238 @@
...
+    });
+  }
+}
+
+function main(node: ESTree.CallExpression, ctx: ProbeContext) {
+  const { sourceFile } = ctx;
+  const { tracer } = sourceFile;
+
+  const algorithm = resolveString(node.arguments.at(0), tracer);
+  if (algorithm === null) {
```
ErwanRaulo: do have to make weak param detection dependent on algothm ? if yes, we lose warnings for algorithm variants.

docs/crypto.weak-argon2.md
clemgbld: idk i feel like we should only accept what is in the recommendation t=2 2 GBI and t=3 64 MIB
https://www.rfc-editor.org/rfc/rfc9106.html#name-recommendations
WDYT ?

workspaces/js-x-ray/src/probes/crypto/resolveStringValue.ts
```diff
@@ -4,7 +4,8 @@ import type { LiteralIdentifier } from "../../VariableTracer.ts";
 
 /**
  * @description
- * Resolves a string literal, or an identifier tracked back to a string literal assignment
+ * Resolves a string literal, or an identifier tracked back to a string literal assignment.
+ * If the identifier is tracked back to a template literal, it will return null.
```
ErwanRaulo, line 8: * Resolves a string literal, or an identifier tracked back to a string literal assignment or a template literal.

docs/crypto.weak-argon2.md
ErwanRaulo: tha's fine, i thought to a more defensive approach to not allow this argon2 variant at all. ok thanks

workspaces/js-x-ray/src/probes/crypto/isWeakArgon2.ts
HoyeongJeon: Thanks! I will check the probe!

docs/crypto.weak-argon2.md
HoyeongJeon: i think both RFC recommendations are for Argon2id, and they don't say much about argon2i. Honestly, I'm not sure what's the better one. Keeping the current passes >= 3, or flagging argon2i as `weak-algorithm` like argon2d. The only thing I'd avoid is replacing the OWASP table with RFC values, cuz that would raise the minimum memory to 64MiB, and `isWeakScrypt` uses the same OWASP table. @ErwanRaulo @clemgbld wdyt?

workspaces/js-x-ray/src/probes/crypto/isWeakArgon2.ts
HoyeongJeon: > do have to make weak param detection dependent on algothm ? if yes, we lose warnings for algorithm variants.

Agree, I'll check the params and algorithm independently

docs/crypto.weak-argon2.md
ErwanRaulo: this kind of detection strenght could be parameterized in the futur. so it's better to keep the analysis less agressive for now, keeping the current passes >= 3.

clemgbld (approved):

fraxken (approved):

PR #708, opened 2026-09-17
fix(js-x-ray): resolve template literal require specifiers

VictorMartins3:
The switch on `arg.type` in `isRequire` covers `Identifier`, `Literal`, `ArrayExpression`, `BinaryExpression` and `CallExpression`. A `TemplateLiteral` falls through to the `default`, which reports `unsafe-import` and drops the dependency.

```js
require("http")   // dependency recorded, no warning
require(`http`)   // unsafe-import, nothing recorded
```

Both are the same string at runtime, so the second one is a false `unsafe-import` and a hole in the dependency list.

Resolved through the existing `toLiteral` helper:

| argument | resolved |
| --- | --- |
| `require("http")` | `"http"` |
| `` require(`http`) `` | `"http"` |
| `` require(`http${suffix}`) `` | `null`, keeps `unsafe-import` |

The last row stays excluded on purpose. With an expression in it the specifier is only known at runtime, which is exactly what the warning is for. `toLiteral` would give `http${0}` there, so the check is `expressions.length === 0` rather than trusting the helper's output.

`isRequire` is the only probe touched, so that is the only behaviour change. Two tests added next to the existing ones, one per row.


jx-probe2.mjs
PierreDemailly, line 1: These files need to be untracked

VictorMartins3: good catch, those were my local benchmark scripts i didnt mean to commit. removed and force pushed, diff is just the probe the test and the changeset now

fraxken (approved):

PierreDemailly (approved):

PierreDemailly:
@allcontributors please add @VictorMartins3 for code

PR #720, opened 2026-09-21
feat(js-x-ray): add weak-pbkdf2 detection probe

### LINKED ISSUE #506: Optional warning for crypto API misuse detection

Vinayak19112003:
Completes the remaining gap of #506 (crypto API misuse detection): weak `crypto.pbkdf2()` / `crypto.pbkdf2Sync()` parameters. Argon2/bcrypt/scrypt and `Math.random()` misuse are already covered by existing probes.

Adds a new optional warning `crypto.weak-pbkdf2` (severity Warning, experimental) that reports:
- `low-iterations` — iteration count below the OWASP minimum (600,000 for SHA-256, 210,000 for SHA-512; unknown digests fall back to the 210,000 floor so only unambiguously weak counts are reported)
- `short-salt` / `hardcoded-salt` — hardcoded string salts (< 16 bytes / >= 16 bytes), same semantics as the scrypt probe

Includes the probe module, warning registration, i18n entries (5 locales), a warning doc page, README catalog entries, a changeset, and 15 tests. Also updated the two `syncIO` pbkdf2 fixtures to use safe parameters, since `isSyncIO` tests run with all optional warnings enabled.

How tested: `npm run build`, `npm run lint`, full `npm run test` (849/849 pass); verified the probe fires on the issue example (`pbkdf2(password, salt, 1000, 64, 'sha512', ...)`) and stays silent on safe usage across require/ESM/`node:crypto` import styles.

Closes #506


fraxken (approved):

workspaces/js-x-ray/src/probes/crypto/isWeakPbkdf2.ts
```diff
@@ -0,0 +1,99 @@
...
+
+const kTracedFunctions = ["crypto.pbkdf2", "crypto.pbkdf2Sync"];
+
+/**
+ * Return the OWASP minimum iteration count for the given digest.
+ * When the digest cannot be resolved statically, fall back to the lowest
+ * recommendation so only unambiguously weak counts are reported.
+ */
+function minIterationsFor(digest: string | null): number {
+  return digest === "sha256" ? kMinIterationsSha256 : kMinIterationsSha512;
```
ErwanRaulo: only special case "sha256" (600,000).
every other statically resolved digest, including "sha1", falls back to the 210,000 floor. 
OWASP recommends 1,300,000 iterations for PBKDF2-HMAC-SHA1. A call like :
```ts 
crypto.pbkdf2(pwd, salt, 300000, 64, "sha1", cb) 
```
would go unreported, even though it's well below the recommendation for that specific digest.

The 210k floor makes sense when the digest can't be resolved statically to avoid false positives on an unknown value, but here the digest is known.

Suggestion: add a dedicated threshold for "sha1" (1,300,000) in `minIterationsFor`

Or is it a deliberate too permissive trade-off ?

ErwanRaulo: `digest === "sha256"` is a strict equality check, but Node/OpenSSL accept digest names regardless of case ("SHA256", "Sha256" work at runtime). 
A call using "SHA256" fails this check, falls back to the 210,000 threshold instead of 600,000, and a weak case (250,000 iterations with SHA-256) would slip through undetected.

Suggestion: normalize with `digest?.toLowerCase() === "sha256"` before comparing.

ErwanRaulo (changes requested):

workspaces/js-x-ray/src/probes/crypto/isWeakPbkdf2.ts
Vinayak19112003: Good catch — done. Added a dedicated 1,300,000 threshold for sha1 in `minIterationsFor`, so a known sha1 digest with e.g. 300,000 iterations is now reported. The 210k floor still only applies when the digest cannot be resolved statically. Added a test for it.

Vinayak19112003: Also fixed — `minIterationsFor` now matches on `digest?.toLowerCase()`, so SHA256/Sha256 get the 600,000 threshold instead of falling back to 210k. Added a regression test (250,000 iterations with 'SHA256' is now flagged).

workspaces/js-x-ray/test/probes/crypto/isWeakPbkdf2.spec.ts
```diff
@@ -0,0 +1,224 @@
...
+    it("should not warn when crypto module is not imported", () => {
+      const code = `
+        const crypto = { pbkdf2() {} };
+        crypto.pbkdf2(password, "short", 1000, 64, 'sha512', (err, key) => {});
+      `;
+      const { warnings: outputWarnings } = analyse(code);
+
+      assert.strictEqual(outputWarnings.length, 0);
+    });
+  });
```
PierreDemailly, line 211: Should we also support PBKDF2 via `globalThis.crypto.subtle` ?
See https://nodejs.org/api/webcrypto.html#deriving-bits-and-keys

I’m fine with leaving this unsupported for now, but maybe we should add a note about it in the docs ?

PierreDemailly (approved):

ErwanRaulo (approved):

fraxken:
@allcontributors please add @Vinayak19112003 for code

